PR #124, opened 2015-06-03
Python3 compatibility

asobolev:
Here are a few changes needed to make nixpy work with python3.

One major change - we cannot use metaprogramming to inject methods as elegant as before (see "The Metaclass Hook in Python 3" here http://python-3-patterns-idioms-test.readthedocs.org/en/latest/Metaprogramming.html). So I used the easy solution and monkey patch classes directly (could be easily changed later if you'd like).

All tests run fine EXCEPT 2 TESTS in data set. Something is wrong with indexing ("selection+offset not within extent"). It would be nice if someone who developed DataSet could take a look and fix these.

Other than that it should be fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Getting Started (Windows)
 
 **Windows Installer**
 
-You can get a windows installer for both, 32- and 64-bit: [nixpy - Beta 1](https://github.com/G-Node/nixpy/releases)
+You can get a [windows installer](https://github.com/G-Node/nixpy/releases) for both, 32- and 64-bit.
 
 **Build NIXPY under Windows**
 
--- a/nix/__init__.py
+++ b/nix/__init__.py
@@ -21,7 +21,7 @@ from nix.data_array import DataSetMixin
 from nix.data_array import DataArrayMixin
 from nix.tag import TagMixin
 from nix.multi_tag import MultiTagMixin
-from nix.entity_with_sources import DataArrySourcesMixin, MultiTagSourcesMixin, \
+from nix.entity_with_sources import DataArraySourcesMixin, MultiTagSourcesMixin, \
     TagSourcesMixin
 
 from nix.section import S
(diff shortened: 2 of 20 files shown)


stoewer:
> could be easily changed later if you'd like

If it is so easy I would prefer to change it right now

asobolev:
I don't have time at the moment to investigate what is the best solution for patching. If you could propose something better than direct patching - I could try to integrate it later (or of course one could always fetch my branch and just do that).

gicmo:
Something strange is going on with coveralls. According to travis coverage is:
```
TOTAL                       1840     36    98%   
```

jgrewe:
Well, this looks better ;)
I have a problem though. The code compiles with python 3.4 but tests are not running on my machine due to some import error  

from nix.core import File, FileMode, Block, DataType, Section, Property, Value, \
ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.54.0: undefined symbol: PyClass_Type

do you have some advice for me?  My system linux mint/ubuntu libboost-python 1.54, used python3.4

gicmo:
@jgrewe that happens when you at runtime you use python3 but it was linked against boost-python for python 2. Maybe the new boost python selection code didn't work correctly or something else went wrong

asobolev:
Jan, could you please send the results of the following with your python3.4:

import platform
platform.system()

as well as

locate libboost_python

Thanks

jgrewe:
sure:

platform.system() just gives 'linux'

Actually it is a linux mint 17, 64 bit 

locate libboost_python:
/usr/lib/x86_64-linux-gnu/libboost_python-py27.a
/usr/lib/x86_64-linux-gnu/libboost_python-py27.so
/usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.54.0
/usr/lib/x86_64-linux-gnu/libboost_python-py34.a
/usr/lib/x86_64-linux-gnu/libboost_python-py34.so
/usr/lib/x86_64-linux-gnu/libboost_python-py34.so.1.54.0
/usr/lib/x86_64-linux-gnu/libboost_python.a
/usr/lib/x86_64-linux-gnu/libboost_python.so

gicmo:
@jgrewe you have the latests commits? Can you make sure it tries and does indeed link against `libboost_python-py34` during `python setup.py build`

asobolev:
How do you build it? 

usually just doing "python3 setup.py build" raises a numpy error (no numpy for python3 by default). So I usually do it with virtualenv, and you?

jgrewe:
sorry was out of the office for a while...
Actually I was behind by a few commits. 
Just trying to build again with: 
sudo python3.4 setup.py build

jgrewe:
build passes with a tiny warning:
In static member function ‘static void* nixpy::PyDataSet::do_export()’:
src/PyDataSet.cpp:285:1: warning: control reaches end of non-void function [-Wreturn-type]
 }

tests are running, installation and import are working now

Issue #122, opened 2015-02-09
simplify exception wrapper


jgrewe:
the previously reported bug of empty exceptions does no longer exist. Thus, the wrapper can again be simplified.

jgrewe:
solved with pr #125

PR #125, opened 2015-06-07
support negative indices in slicing...

jgrewe:
plus re-factoring of the exceptions which no longer needed the formerly required checks


gicmo:
Commit 13a5db1 is related to #122

PR #127, opened 2015-07-27
Fix small py3 related build warning

gicmo:
As @jgrewe detected in pr #124 there is a small build warning
since we don't actually return a value in PyDataSet::do_export
for py3.


gicmo:
Heh coveralls. Didn't even touch python source files.

jgrewe:
There is something really weird going on with coveralls :)

Issue #81, opened 2014-10-21
Installation


jgrewe:

stoewer:
Does someone know what this issue was about?

jgrewe:
the only thing that I can imagine are the installation instructions in the documentation. But honestly, I have no clue

jgrewe:
Actually, I am by now pretty sure, it was related to the documentation because it was opened the same day, the tutorial issue was created.

stoewer:
> Actually, I am by now pretty sure, it was related to the documentation because it was opened the same day, the tutorial issue was created.

OK, this makes sense. Thanks.

jgrewe:
solved with pr #135

PR #135, opened 2015-12-03
Installation guide

stoewer:
Fixes issue #81

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -2,19 +2,96 @@
 Installation Guidelines
 =======================
 
-TODO write something
+Install NIX and other dependencies
+==================================
 
-Install Dependencies
-====================
+Because NIXPy is build upon the `NIX library`_ it is necessary to install the NIX library before you can install or build NIXPy on your system.
+The following two sections explain how NIX can be installed on Windows and Linux.
 
-TODO write something
+.. _NIX library: https://github.com/G-Node/nix
 
-Build and Install NIX
-=====================
+Linux
+-----
 
-TODO write something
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -2,19 +2,96 @@
...
+If the PPA was not already added to your system, you can do so by executing the following commands:::
+
+    sudo add-apt-repository ppa:gnode/nix
+    sudo apt-get update
+
+Once the PPA was added NIXPy can be installed via apt-get::
+
+    sudo apt-get install python-nix
+
+If you want to use NIXPy on other distributions, pleas follow the instructions for building NIXPy from source.
```
jgrewe: add a (see below) at the end of line 77?

docs/source/install.rst
```diff
@@ -2,19 +2,96 @@
...
+    sudo add-apt-repository ppa:gnode/nix
+    sudo apt-get update
+
+Afterwards you can use apt to install the NIX package::
+
+    sudo apt-get install libnix-dev
+
+If you want to use NIX on other distributions you have to compile and install NIX from source.
+
+.. _PPA: https://launchpad.net/~gnode/+archive/ubuntu/nix
```
jgrewe: I do  not understand how the ppa would help to build it from source? Should this link be somewhere closer to line 15?

stoewer: This an external link to our PPA which is used in line 16.  I chose to define all links and references at the end of the section where they are used for the first time. To me this seems to be a good compromise between having them right there where they are used (which may impair the reading flow) and putting them at the end of of the file (where one might not see them).

Issue #137, opened 2015-12-04
Writing of data via index+slices silently ignores data exceeding the DataArray's shape


gicmo:
Look here:

```python
import nix
fd = nix.File.open('test.h5', nix.FileMode.Overwrite)
b = fd.create_block('b', 'notype')
d = b.create_data_array('da', 'stillnotype')
d = b.create_data_array('da', 'stillnotype', nix.DataType.Double, (0,0))
d[:, :] = np.ones((23, 42))
d[...]
# Out: array([], shape=(0, 0), dtype=float64)
d2 = b.create_data_array('da', 'stillnotype', nix.DataType.Double, (12,13))
d2 = b.create_data_array('da2', 'stillnotype', nix.DataType.Double, (12,13))
d2[:, :] = np.ones((23, 42))
d2[...]
d2.shape
# Out: (12, 13)
```

PR #142, opened 2015-12-06
Support auto-extending for sliced, index-based data writing

gicmo:
Should fix bug #137


gicmo:
Needs more thinking how to do handle the writing part.

PR #146, opened 2016-02-01
[RFC] Rename nix to nixpy

souravsingh:


gicmo:
I am afraid this is opening the "box of pandora" since there is still no consensus how we should handle the name clashes. See issue G-Node/nix#595

PR #151, opened 2016-02-20
Update README with OS X Guide and fix typos

sudoankit:
Made a section for OS X.
My fork passes travis successfully. [Logs are here.](https://travis-ci.org/sudoankit/nixpy/jobs/110552142)
(It's text anyways.)

Update to travis.yml bit delayed due to minor hiccups arising from different OS X versions. 
Will do it soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,14 @@ In order to build the NIXPY library a compiler is needed (gcc, clang). Further n
 
 _Instructions_
 
-```bash
+```
+bash
 # install dependencies
 sudo apt-get install libboost-python-dev build-essential python-dev python-setuptools python-numpy
 
 # clone NIX
 git clone https://github.com/G-Node/nixpy
-cd nix
+cd nixpy
 
 # build nixpy
 python setup.py build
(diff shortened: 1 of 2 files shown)


jgrewe:
Hi @sudoankit, thanks for your pull request!
Could you rebase your branch and clean the commit history that it is not a back-and-forth.

README.md
```diff
@@ -63,6 +64,62 @@ python setup.py test
...
+
+Get the source code from GitHub using :
+
+`git clone https://github.com/G-Node/nixpy.git`
+
+**Build NIXPY under OS X**
+
+_Dependecies_
+
+- NIXPY requires installation of [NIX](https://github.com/G-Node/nix). Make sure you install NIX first.
```
jgrewe: One could mention how to install nix via homebrew

README.md
```diff
@@ -63,6 +64,62 @@ python setup.py test
...
+
+OS X default python works well.
+
+_Instruction to build_
+
+```
+bash
+
+# install dependencies if required
+sudo apt-get install libboost-python-dev build-essential python-dev python-setuptools python-numpy
```
jgrewe: you sure, that apt-get will work on OS X?

README.md
```diff
@@ -63,6 +64,62 @@ python setup.py test
...
+
+# build
+python setup.py build
+
+# unit tests
+python setup.py test
+
+# install 
+sudo python setup.py install
```
jgrewe: I am not sure, if we should suggest to use sudo on OS X. It will interfere with the system installed version of python. I think this is rather tricky since it depends strongly on how python was installed via home-brew, the system one, or via macports ...
@gicmo  Is there a "unified" way how to do it?

README.md
sudoankit: Opps. I am extremely sorry to have overlooked this.

README.md
sudoankit: Yes, I will add a line.

README.md
sudoankit: - [Homebrew]  installed python would be fine without sudo.
- [Macports] I wouldn't suggest

We could suggest to install anaconda and then install nixpy. `conda` automatically changes the `.bashrc`.
And, the new version of OS X, has a feature called `rootless`, which doesn't allow critical system files to be modified even with `sudo` .

sudoankit:
Please close the pull request and making a new one, for the `Readme.md` only.
This has become a zombie PR, now as I deleted the fork, forked it again.
I learned about rebasing and cleaning now, so now I wil do it properly.
Should I add `sudo` for installation, or leave it as it is?

jgrewe:
I would rather leave the sudo away.

PR #152, opened 2016-02-29
Upgrade travis to trusty and add a cmake support

gicmo:
The cmake support is mainly for IDE support (CLion 1.5 has python support).
Travis upgrade to make build (start) faster.


gicmo:
PR #153 should be merged first, I will rebase it then accordingly.

PR #159, opened 2016-03-06
fixed python code block in overview.rst

Prabhnith:
-fixed typos
- `tag.references.add(data)` throws AttributeError , fixed with `tag.references.extend(data)` 
- `np.rando.randn` requires integer ,so fixed
-`np.zeros((3,2))` required

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -248,8 +248,8 @@ Let's say we want to give each  point a name, we can create a feature like this:
 
 .. code-block:: python
 
-		spot_names.append_set_dimension()
 		spot_names = block.create_data_array('spot ids', 'nix.feature',dtype=nix.DataType.Int8, data=[1, 2])
+		spot_names.append_set_dimension()
 		feature = tag.create_feature(spot_names, nix.LinkType.Indexed)
 
 We could also say that each point in the tagged data (e.g. a matrix of
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -195,7 +195,7 @@ dimensionality of the referenced data.
 		tag = block.create_tag('interesting part', 'nix.roi', position)
 		tag.extent = extent
 		# finally, add the referenced data to this tag
-		tag.references.extend(data)
+		tag.references.append(data)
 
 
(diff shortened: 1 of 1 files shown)


docs/source/overview.rst
```diff
@@ -248,8 +248,8 @@ Let's say we want to give each  point a name, we can create a feature like this:
 
 .. code-block:: python
 
-		spot_names = block.create_data_array('spot ids', 'nix.feature', data=['a', 'b'])
 		spot_names.append_set_dimension()
+		spot_names = block.create_data_array('spot ids', 'nix.feature',dtype=nix.DataType.Int8, data=[1, 2])
```
jgrewe: this would not work: you cannot append a dimension before the DataArray exists. Please invert the order of lines 251 and 252.

Could you also change the introductory sentence to: "Let's say we want to give each point an ID, we can create a feature like this:"

docs/source/overview.rst
```diff
@@ -195,7 +195,7 @@ dimensionality of the referenced data.
 		tag = block.create_tag('interesting part', 'nix.roi', position)
 		tag.extent = extent
 		# finally, add the referenced data to this tag
-		tag.references.add(data)
+		tag.references.extend(data)
```
jgrewe: not sure right now, but this should be ```append ```. ```extend``` should also work but for single entries append might be more appropriate

jgrewe:
Thanks for the typo fixes!

docs/source/overview.rst
Prabhnith: append is for single key and extend is for multiple keys.
I think data should be given to extend() but both will work.

jgrewe: it still should be append :)
try 
````
x = []
x.extend(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-b0aa5471eb38> in <module>()
----> 1 x.extend(1)

TypeError: 'int' object is not utterable

x.append(1)  # works
x.extend([1])  # word as well

````

Prabhnith:
ok got it ! :)

PR #160, opened 2016-03-06
Update overview.rst

Prabhnith:


docs/source/overview.rst
```diff
@@ -256,8 +256,8 @@ We could also say that each point in the tagged data (e.g. a matrix of
 measurements) has a corresponding point in an input matrix.
 
 .. code-block:: python
-		
-		input_matrix = np.random.randn(data.shape)
+		a,b=data.shape
+		input_matrix = np.random.randn(a,b)
```
jgrewe: actually I would prefer the previous version since it is more compact and also illustrates interactions better

docs/source/overview.rst
```diff
@@ -298,7 +298,7 @@ same brain region of the same animal. To represent this hierarchy,
 
 .. code-block:: python
 
-		subject.block.create_source('subject A', 'nix.experimental_subject')
+		subject=block.create_source('subject A', 'nix.experimental_subject')
```
jgrewe: would you mind, adding white-spaces around the =? Thanks,

Prabhnith:
ok , i will add space.

but sir for the case of
I have changed this
`input_matrix = np.random.randn(data.shape)`
to
`a,b=data.shape
matrix = np.random.randn(a,b)`

because
np.randon.randn() accepts integer range as parameter and the data.shape is a tuple and giving an error.
so by assigning a,b=data.shape we can attain it without throwing any error.

jgrewe:
ups, yes, you are right. Would you mind to switch to ```np.random.random``` which does accept a tuple... Thanks.

Prabhnith:
ok done, changed

PR #161, opened 2016-03-06
written unit code example in Overview.rst

Prabhnith:


docs/source/overview.rst
```diff
@@ -353,4 +353,17 @@ will handle unit scaling. The correct data will be returned even if
...
+		
+		#single SI unit is supported like mV,cm etc.
+		tag.units="cm"
+		
+		#for compound units we can make list
+		tag.units=["mV","cm"]
+		
+		#we can make a compound_unit attribute to get the compound units
+		#result will show : mV/cm
+		tag.compound_unit = "/".join(tag.units)
```
jgrewe: even though this will not raise an exception the created compound unit won't be persisted in the file!

docs/source/overview.rst
```diff
@@ -353,4 +353,17 @@ will handle unit scaling. The correct data will be returned even if
...
-TODO: example code
+.. code-block:: python
+
+		x_positions=[2,4,6,8,10,12]
+		tag=block.create_tag('unit example','nix.sampled',x_positions)
+		
+		#single SI unit is supported like mV,cm etc.
+		tag.units="cm"
+		
+		#for compound units we can make list
```
jgrewe: There is a misunderstanding here:
tags can point into multi-dimensional data. Thus they do always require one unit per dimension of the data they are pointing at (if the data dimensions are not dimensionless, uh, i.e. have no unit). Hence, we always need to set them as a list (```tag.units = ["ms", "cm"]``` if the data has a time-dimension in the first and a spatial dimension in the second). If there is a single unit that is a compound of two SI units (mV/cm) you should be able to give them with ```tag.units = ["mV/cm"]```.

docs/source/overview.rst
```diff
@@ -353,4 +353,17 @@ will handle unit scaling. The correct data will be returned even if
...
 
 
-TODO: example code
+.. code-block:: python
+
+		x_positions=[2,4,6,8,10,12]
+		tag=block.create_tag('unit example','nix.sampled',x_positions)
+		
+		#single SI unit is supported like mV,cm etc.
+		tag.units="cm"
```
jgrewe: actually, this line (362) won't work, at least not in my version :)
The error message, however, is really bad. Thanks for pointing that out..

Prabhnith:
by using this we can get the compound units
selection_019

PR #163, opened 2016-03-08
removed blank spaces and added lines according to PEP

Prabhnith:


jgrewe:
I guess there is more to come :)
I'd prefer, if you would add further small fixes (PEP, typo) to this pull request. Just continue committing to this branch on your fork.
If this is final, let me know and I will merge.

Prabhnith:
done. any reviews @jgrewe

docs/source/examples/multipleROIs.py
```diff
@@ -88,30 +90,33 @@ def plot_roi_data(tag):
...
+    positions = block.create_data_array(
+        "ROI positions", "nix.positions", data=roi_starts)
+    positions.append_set_dimension()  # these can be empty
     positions.append_set_dimension()
-    
+
     # create the extents DataArray
-    extents = block.create_data_array("ROI extents", "nix.extents", data=roi_extents)
+    extents = block.create_data_array(
+        "ROI extents", "nix.extents", data=roi_extents)
```
jgrewe: As I see it PEP is a suggestion and this line (and 106, 107) are examples were to bend the rules :) this is quite ugly...

If you need to wrap the line, do it before the last argument respectively the last arguments that are beyond the 80 character threshold.

Prabhnith:
But when we follow threshold , the readability of the code is decreased. 
So one liner code is good.
and 106 ,107 are not much of violations as they are 84 in characters , we can take it :D 
ok, to change the code from ugly to nice , i will make it one liner

PR #168, opened 2016-05-20
Backend selection and pure Python Value class

achilleas-k:
Laying the foundation for supporting dual backends.

Changes:
- All Mxin classes subclass `object` instead of the base class they are patching. This will allow us to patch both C++ and Python classes with the same Mixins. The `inject` method checks the `base` argument for the cases where a class is not loaded (when the HDF5 backend is unavailable).
- `nixio.File.open` accepts `backend` argument which defaults to `"hdf5"` for now. Raises an error if HDF5 is selected and the C++ backends aren't available. The argument returns a `File` object from the appropriate backend.
- `nixio.FileMode` is now a Python class with 3 static members for the 3 file modes. PyFile.cpp is responsible for mapping the mode characters used in Python to the appropriate `FileMode` enum value.
- All TestCases have been renamed to have leading underscores so that they act as base classes and are ignored by test discovery. Each TestCase class is subclassed twice, once for each backend. The backend is defined in the `setUp` method. C++ tests are skipped when the backend is unavailable. Currently, all Python tests are overridden with `pass` until the classes are implemented.
- `nixio.Value` and `nixio.DataType` are now pure Python classes. PyValue.cpp has been completely removed and the `DataType` enum is no longer exported to Python.
  - The new `Value` class, defined in `nixio/value.py`, is converted to a `nix::Value` type directly, using the new converter defined in PyProperty.cpp (`value_transmogrify`).
  - The Python `DataType` class simply aliases numpy dtypes (e.g., `np.int32`). The `pyDtypeToNixDtype` function has a new case `S` for strings.


nixio/test/test_file.py
```diff
@@ -81,3 +87,30 @@ def test_file_find_sections(self):
         assert(len(self.file.find_sections(limit=1)) == 2)
         assert(len(self.file.find_sections(filtr=lambda x : "level2-p1-s" in x.name)) == 2)
         assert(len(self.file.find_sections(filtr=lambda x : "level2-p1-s" in x.name, limit=1)) == 0)
+
+
```
stoewer: Shouldn't the `@unittest.skipIf` be here instead of `_FileTest`?

nixio/test/test_property.py
```diff
@@ -171,3 +170,53 @@ def test_value_attrs(self):
...
         value.uncertainty = 0.5
         assert(value.uncertainty == 0.5)
+
+
+@unittest.skipIf(skip_cpp, "HDF5 backend not available.")
+class TestPropertyCPP(_TestProperty):
+
+    def setUp(self):
+        self.file    = File.open("unittest.h5", FileMode.Overwrite,
+                                 backend="hdf5")
```
stoewer: Everything below can be implemented in a `setUp` method provided by the super class.

nixio/test/test_file.py
achilleas-k: This was a mistake. Fixed it now.

nixio/test/test_property.py
achilleas-k: I was also thinking that the test superclasses could define the setUp method and the subclasses could just define their own backend in a static variable, used in the superclass setUp method.

stoewer: Sounds fine. Do you want to include the change in this PR or do you plan to do it later?

achilleas-k: Done.

achilleas-k: Pushed the changes now. Seems to be some delay in them appearing here.

PR #169, opened 2016-05-24
Python Block class

achilleas-k:
# Pure python Block class.

Adding Block class to pycore. The entity class hierarchy is based on the NIX backend hierarchy, with appropriate modifications.

New entities are created in the HDF5 file using the `_create_new` methods. The constructor is used to initialise a Python object from an existing file object. The `_create_new` method first creates the object in the HDF5 file and then calls the constructor to initialise and return the Python object.

Block tests have been modified in the same way as the File tests to test both backends.

Placeholder method headers have also been created for all classes.


nixio/pycore/block.py
```diff
@@ -0,0 +1,86 @@
...
+class Block(EntityWithMetadata):
+
+    def __init__(self, h5obj):
+        super(Block, self).__init__(h5obj)
+        # TODO: Validation for containers
+
+    @classmethod
+    def _create_new(cls, parent, name, type_):
+        newentity = super(Block, cls)._create_new(parent, name, type_)
+        # TODO: Create child containers
```
stoewer: In the NIX C++ library we create such subgroups only on demand. But it is not a format violation to create empty ones. 

It doesn't has to be implemented now, but it would be cool to have this feature in the pure python implementation too. You should either open an issue for this feature or implement something like the [optGroup](https://github.com/G-Node/nix/blob/master/backend/hdf5/h5x/H5Group.hpp#L237) now.

achilleas-k: I've created issue #170 to handle this so as to not hold back this PR.

PR #171, opened 2016-05-27
Python Group and Section classes

achilleas-k:
Adding Group and Section classes to pycore.

Entity classes have been reorganised a bit to solve issues with import cycles.

TestGroup and TestSection tests all pass, except `test_section_properties` which fails since the Property class isn't implemented yet.
Section tests in TestFile have been enabled again and are passing.

### DIFF
--- a/nixio/pycore/entity.py
+++ b/nixio/pycore/entity.py
@@ -8,7 +8,6 @@
 
 from time import time
 from . import util
-from . import exceptions
 
 
 class Entity(object):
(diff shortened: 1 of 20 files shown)


nixio/pycore/entity.py
```diff
@@ -67,31 +66,8 @@ def __str__(self):
     def __repr__(self):
         return self.__str__()
 
-util.create_h5props(NamedEntity, ("name", "type", "definition",
-                                  "id", "created_at", "updated_at"))
+util.create_h5props(NamedEntity,
```
stoewer: The signature of `create_h5props` with a set of names and a set of types is a bit ugly. Just use a dict instead:

```python
util.create_h5props(NamedEntity, {"name": str, "type": str, "definition": str})
```

Or you could implement `create_h5props` using **kwargs:

```python
util.create_h5props(NamedEntity, name: str, type: str, definition: str)
```

achilleas-k: I agree, it's a awkward in its current state. Will modify as suggested.

nixio/pycore/section.py
```diff
@@ -17,43 +20,64 @@ def __init__(self, h5obj):
...
+    def create_section(self, name, type_):
+        util.check_entity_name_and_type(name, type_)
+        sections = self._h5obj["sections"]
+        if name in sections:
+            raise exceptions.DuplicateName("create_section")
+        sec = Section._create_new(sections, name, type_)
+        return sec
+
+    @property
+    def parent(self):
```
stoewer: If a Section was created from a group opened via the `metadata` hard link of an `EntityWithMetadata` the parent of that group might not be the group of the parent section but the entity providing the hard link instead.

PR #172, opened 2016-05-27
Python DataArray class

achilleas-k:
Python classes for DataArray, DataSet, and Dimensions.

Tests enabled for both backends for DataArray.
Dimension tests are bypassed for now.

This PR also adds python implementations of all the utility functions that are exposed from nix.

### DIFF
--- a/nixio/pycore/entity_with_metadata.py
+++ b/nixio/pycore/entity_with_metadata.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2016, German Neuroinformatics Node (G-Node)
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted under the terms of the BSD License. See
+# LICENSE file in the root of the Project.
+
+from .entity import NamedEntity
+from .section import Section
+
+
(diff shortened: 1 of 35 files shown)


achilleas-k:
New commit for DimensionType.
The C++ enumerator for DimensionType is no longer exposed to Python and the `dimension_type` property is removed from the classes in PyDimensions.cpp.
With this change, only a Python version of DimensionType exists in nixio.

Three new Mixin classes are also added to nixio/data_array.py, one for each Dimension class. Each Dimension class (C++ backend only) is patched with a static property called `dimension_type`.

nixio/pycore/entity_with_metadata.py
```diff
@@ -0,0 +1,27 @@
...
+        # TODO: Additional validation for metadata
+
+    @classmethod
+    def _create_new(cls, parent, name, type_):
+        newentity = super(EntityWithMetadata, cls)._create_new(parent,
+                                                               name, type_)
+        return newentity
+
+    @property
+    def metadata(self):
```
stoewer: As mentioned in my comment on #171 the following will most likely not work (with respect to the `parent` property of the `Section` class.

PR #175, opened 2016-06-10
Container and property method refactoring

achilleas-k:
Major refactoring of Pycore/H5Py backend. Runtime patching for container methods (get, delete, count, etc) and property methods (setters and getters) is no longer performed.

## New H5Group class

H5Group is a wrapper class which pycore classes now use to interact with h5py.Group objects. The class implements methods such as `get_by_id`, `get_by_pos`, and `delete`, which the nix classes call to access child objects. Similarly, it implements `set_attr` and `get_attr` for manipulating Group attributes in the HDF5 file.

#### Type checking

Some basic type checking occurs when setting HDF5 attributes. Type checking is the responsibility of each NIX class, while the `set_attr` method of `H5Group` simply handles the writing of the values to the backend.

## Mixins

Each nix pycore class subclasses its respective Mixin class, instead of using the inject method as is done for the C++ backend. To make the separation between the backends cleaner, the inject method is called by the top level `__init__.py` when the C++ backend is available and all injections occur inside the `util/inject.py` module.


achilleas-k:
The last commit is a workaround for a bug I've been investigating the last two days.
I'm not sure if it's best to merge the PR and find the root cause later, or hold off and solve the issue first.

PR #178, opened 2016-06-20
PyCore: Tags, MultiTags, Sections, Properties

achilleas-k:
This PR makes PyCore feature complete. All tests pass successfully.

### New classes
The changes include full implementations of the following classes:
- BaseTag (Tag and MultiTag base class)
- Tag
- MultiTag
- Source
- Property
- DataView
- Feature

### Changes and fixes
The following classes have also had changes made to accommodate the new classes and fix some incorrect functionality:
- Section: Creating and getting properties and creating links to other sections.
- DataSet: The base class for DataArray, previously removed. Introduced again to act as a base class for both DataArray and DataView classes.
- Dimensions: Bug fixes with property read/write.
- H5Group: Methods for reading and writing to contained data object (H5DataSets) and utility functions for data tree traversal (getting root objects and finding children).

I also made small changes to PyTag.cpp and PyMultiTag.cpp to explicitly convert from strings to C++ LinkType objects, since the C++ LinkType class is no longer exposed in nixpy.

I removed the wrapper for the C++ implementations of util.names and util.units. These have been completely rewritten for nixpy and no longer need to be exposed. The utility tests now only run for pycore.util functions.

###  HDF5 order tracking and indexing
File and H5Group now use the low level HDF5 API when creating files and groups respectively to enable order tracking and indexing, preserving the creation order for objects in the file.


nixio/pycore/h5group.py
```diff
@@ -70,34 +94,76 @@ def get_dataset(self, name):
...
-        if the a dataset of that name does not exist in the Group.
+        Returns the data contained in the dataset identified by 'name', or an
+        empty list if a dataset of that name does not exist in the Group.
 
         :param name: The name of the dataset
         :return: The data contained in the dataset as a numpy array or None
         """
         if name not in self.group:
-            return tuple()
+            return []
```
stoewer: Isn't it better to raise an error here?

achilleas-k: The same way None is returned when an attribute that doesn't exist is requested in `get_attr`, I thought it was more appropriate to return an empty list here.
Since certain data properties are removed when they're empty, such as SetDimension.labels, I felt it should be the calling class' responsibility to allow empty or missing data.

I think NIX behaves the same way. For instance, in NIX, RangeDimension.ticks() handles the error if the group does not contain a "ticks" or "data" data object. Conversely, SetDimension.labels() returns an empty vector if "labels" is missing.

stoewer: OK

PR #184, opened 2016-07-20
Creation order indexing

achilleas-k:
Now using creation order when accessing elements by index, same as G-Node/nix#618.

PR also includes a number of bug fixes.


achilleas-k:
Python 2 tests failing. I think it's something to do with string types (again).

achilleas-k:
Fixed. Empty property creation was being initialised with Value(0) instead of leaving Property.values empty.

Issue #173, opened 2016-05-28
[test] Write then read with different backend


achilleas-k:
Once the Python H5Py backend is complete, we will need new tests for validating backend compatibility.
NIX files should be created using one backend and then read in with the other, testing that the structure of the data is valid and unchanged.

PR #185, opened 2016-07-26
[PyCore] Backend compatibility tests

achilleas-k:
## Full backend compatibility tests

This PR includes tests where a file is created with one backend (hdf5 or h5py) and is then read by the other to validate that both backends behave the same way.

All other changes are fixes for bugs which were discovered by the new tests.


achilleas-k:
Closes #173.

Issue #200, opened 2016-08-11
[setup] Flag for explicitly skipping C++ backend


achilleas-k:
The setup script should have an option for skipping the compilation of the C++ NIX bindings, even when NIX is available.

Issue #199, opened 2016-08-11
[setup] Ignore native extension when NIX is not available


achilleas-k:
The setup script should not fail when NIX is not installed. It should simply install nixpy for use with pycore (h5py) backend.

PR #206, opened 2016-08-30
Python only builds

achilleas-k:
Setup script updated to support building without the NIX dependency.

The setup script detects if NIX is available by attempting to compile a dummy C++ file that simply imports `nix.hpp` and checking if it fails (see new file `checknix.py`).

Additionally, Python only builds can now be explicitly performed using `python setup.py build --pyonly`.

Travis build config now includes 4 configurations: Python 2.7 and 3.4 with and without NIX.

### DIFF
--- a/checknix.py
+++ b/checknix.py
@@ -1,0 +1,57 @@
+import os, sys
+from textwrap import dedent
+import tempfile
+
+import distutils.sysconfig
+import distutils.ccompiler
+from distutils.errors import CompileError, LinkError
+
+
+def check_nix(libdirs=(), incdirs=()):
+    """
+    Check if NIX is available by trying to compile a tiny program.
+    """
+
+    cpp_code = dedent("""
+    #include <nix.hpp>
+
+    int main() {
+        return 0;
+    }
+    """)
+    with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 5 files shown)


achilleas-k:
Looks like it's not detecting NIX properly. Fixing now.

achilleas-k:
Fixed.

achilleas-k:
Relevant issues: #199, #200.

checknix.py
```diff
@@ -0,0 +1,64 @@
...
+    compiler.include_dirs.extend(incdirs)
+    distutils.sysconfig.customize_compiler(compiler)
+
+    stderr = os.dup(sys.stderr.fileno())
+    errfile = open(os.path.join(tmpdir, "check_nix.err"), 'w')
+    # redirect stderr to file
+    os.dup2(errfile.fileno(), sys.stderr.fileno())
+    try:
+        compiler.compile([file_name], output_dir=tmpdir,
+                         extra_postargs=["--std=c++11"])
```
gicmo: NIX (headers) also depend on boost (headers) so if boost is installed in a non-standard location then nix will fail to build. Could work around that by also passing the boost incdirs.

achilleas-k: I'll pass those as well.

PR #217, opened 2016-10-07
Windows: Add wheel and appveyor support

matham:
Regarding wheels, this works by getting the most recent nix (currently the version is set using `NIX_VERSION` in appveyor) binaries and including them in the wheel. It adds them under share/nixio/bin in your python installation folder. Then at runtime that folder, if it exists, is added to the PATH.

This is how we do it at [kivy ](https://github.com/kivy/kivy) and is similarly how numpy [does it](https://github.com/numpy/numpy/issues/6923) (although a bit differently). This makes it easy to install and uninstall since you just need to do pip install wheel_name, provided you have the numpy and h5py dependencies installed of course. If it's uploaded to pypi, then they just do `pip install nixio`.

When packaging and building a end user exe using e.g. pyinstaller, for projects that use nixio, this is also super easy, as the project just specifies that pyinstaller include the contents of that folder which includes al the binaries.

Currently, it works with 64 and 32 bit versions of py 2.7, 3.4, and 3.5. The wheels are found under the [artifacts tab](https://ci.appveyor.com/project/matham/nixpy/build/1.0.52) for each build.

For you to make it work, you'd need to merge this PR and then in appveyor you'd import the nixpy project in your account and in the settings tell it it to build all the branches, but only the branches that include an appveyor.yml file. The settings that run the builds will then be provided by the appveyor.yml file included in the PR (most of the setting in the general tab in appveyor control panel cannot be controlled from the appveyor.yml file so you have to set them there).

Finally, the docs need to be updated, but that can be done after we're sure it works. Also, I tried to set up the tests to run on appveyor for each build, however, the h5py and numpy wheels, when using vanilla python, must be downloaded from www.lfd.uci.edu/~gohlke/pythonlibs/ which cannot be done from the command line. So I couldn't run it. We could download manually and then upload those wheels somewhere if you have a server, in which case it should be easy to setup the tests to run automatically. Finally, appveyor runs and generates wheels for every commit, PR, and git tag. So automatic tests would be useful there.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,158 @@
+version: 1.0.{build}
+environment:
+  NIX_VERSION: 1.2.0
+  NIX_BUILD_DIR: C:\nix_build
+  DEPLOY_DIR: deploy
+  BOOST_VERSION: 1.56.0
+  BOOST_VERSION_SHORT: 1_56
+  matrix:
+  - PYVER: 27
+    BITTNESS: 64
+  - PYVER: 27
+    BITTNESS: 32
+  - PYVER: 34
+    BITTNESS: 64
+  - PYVER: 34
+    BITTNESS: 32
+  - PYVER: 35
+    BITTNESS: 64
+  - PYVER: 35
+    BITTNESS: 32
+build_script:
+- ps: >-
+    function Check-Error
+
+    {
+      param([int]$SuccessVal = 0)
+      if ($SuccessVal -ne $LastExitCode) {
+        throw "Failed with exit code $LastExitCode"
+      }
+    }
+
+
+    $old_path = "$env:PYTHONPATH"
+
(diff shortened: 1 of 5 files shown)


matham:
Just noticed that it's failing on py3.5. That's because the compiler for [3.5 changed (and now uses ucrt)](http://stevedower.id.au/blog/building-for-python-3-5/), but we're still using compiler version 12 so the arguments passed by python is wrong. We could try using the proper compiler for python 3.5+ since it's > v12 which seems to be the minimum visual c version required for nix.

achilleas-k:
I've set up nixpy on Appveyor and I've been trying to get it to run the build for this PR but it seems like it's not possible to just trigger a build on an existing PR without any action.

I'm going to merge this and fix any build issues in master.

Issue #218, opened 2016-10-18
Build broken on OSX due to boost python linker error


gicmo:
Command line & error:
```
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ld: library not found for -llibboost_python-mt
```
```
clang++ -bundle -undefined dynamic_lookup -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk build/temp.macosx-10.12-x86_64-2.7/src/core.o build/temp.macosx-10.12-x86_64-2.7/src/docstrings.o build/temp.macosx-10.12-x86_64-2.7/src/PyExceptions.o build/temp.macosx-10.12-x86_64-2.7/src/PyBlock.o build/temp.macosx-10.12-x86_64-2.7/src/PyFile.o build/temp.macosx-10.12-x86_64-2.7/src/PySection.o build/temp.macosx-10.12-x86_64-2.7/src/PyProperty.o build/temp.macosx-10.12-x86_64-2.7/src/PySource.o build/temp.macosx-10.12-x86_64-2.7/src/PyDataArray.o build/temp.macosx-10.12-x86_64-2.7/src/PyDataSet.o build/temp.macosx-10.12-x86_64-2.7/src/PyDimensions.o build/temp.macosx-10.12-x86_64-2.7/src/PyFeature.o build/temp.macosx-10.12-x86_64-2.7/src/PyTag.o build/temp.macosx-10.12-x86_64-2.7/src/PyMultiTag.o build/temp.macosx-10.12-x86_64-2.7/src/PyResult.o build/temp.macosx-10.12-x86_64-2.7/src/PyGroup.o -L/usr/local/lib -L/usr/local/lib -L/usr/local/opt/openssl/lib -L/usr/local/opt/sqlite/lib -L/usr/local/lib -L/usr/local/lib -lnix -llibboost_python-mt -o build/lib.macosx-10.12-x86_64-2.7/nixio/core.so
```

I suspect PR #217, i.e. commit fd7d0f0

jgrewe:
fixed via #220

PR #220, opened 2016-10-18
findboost: detect win lib via ext == ".lib"

gicmo:
This will fix the macOS compile bug (issue #218). Hopefully not break win or GNU/Linux.
I actually haven't verified the other changes that were done with PR #217 just made sure
it is working on macOS again.


gicmo:
I have enable appveyor so we can PRs agains the windows env too, here (on this PR) esp. useful since I was changing a build related thing.

PR #222, opened 2016-10-20
Adds bools to valid_types

klanderson:
This change is necessary for [Neo](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/nixio.py) to save a nix file. Specific use case was saving a Neo block loaded from a Blackrock .nev file.

```
data_block = neo.io.BlackrockIO(file_name, verbose=False).read_block(name=file_name,
                                                                   nsx_to_load=None,
                                                                   channels='all',
                                                                   units='all',
                                                                   load_waveforms=False,
                                                                   load_events=True,
                                                                   lazy=False,
                                                                   cascade=True)

writer = neo.io.NixIO(filename='b.nix', mode="ow")
writer.write_block(data_block)
```

### DIFF
--- a/nixio/value.py
+++ b/nixio/value.py
@@ -18,7 +18,7 @@ except NameError:
 
 bools = (bool, np.bool_)
 
-valid_types = (Number, strings)
+valid_types = (Number, strings, bools)
 
 
 class DataType(object):


achilleas-k:
Thank you for the PR.

I intend to merge this PR with a warning:
Bools were disabled in nixpy metadata because of an incompatibility with the way nix stores booleans. I was holding off enabling them until the compatibility was fixed (coming very soon). Since there seems to be a requirement for this, and it works when using nixpy in pure Python mode (which is feature complete), I think it's appropriate to enable it in master.

PR #225, opened 2016-10-27
Release instructions/checklist

achilleas-k:
Release instructions for nixpy.

### DIFF
--- a/release_instructions.md
+++ b/release_instructions.md
@@ -1,8 +1,8 @@
 # Checklist for NIXPY release
 
 1. Create branch (if necessary)  
-   For major releases (e.g., 1.2.0, 1.3.0, etc) create a new branch called `vX.Y` (e.g., `v1.2`).
-   This is not necessary for point releases (e.g., `1.2.0 → 1.2.1`)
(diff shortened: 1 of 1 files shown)


release_instructions.md
```diff
@@ -0,0 +1,20 @@
+# Checklist for NIXPY release
+
+1. Create branch (if necessary)  
+   For major releases (e.g., 1.2.0, 1.3.0, etc) create a new branch called `vX.Y` (e.g., `v1.2`).
```
cgars: I am not sure whether  and major and poinr relases are good name choices (yes i am nitpicking)

achilleas-k:
I agree. I'll see if I can use less ambiguous terms.

jgrewe:
Should we also copy the instructions for the debian packages to this document (https://github.com/G-Node/nix/blob/master/docs/release_instructions.md). Or, more generally, do we want to keep up providing debian packages, via launchpad?

achilleas-k:
That's a good question. Is having a deb package more convenient for Debian or Ubuntu users than using pip? I suppose we could automate at least part of it.

In fact, I'm thinking now we could automate significant parts of it.

I'll see if I can draft a release script and add it to the PR.

PR #233, opened 2016-11-04
Version compatibility checks

achilleas-k:
Counterpart to G-Node/nix#638.

Compares supported file version with the format version of the file being opened and decides whether it can be opened in the requested mode (read only or read write).

Follows the [NIX documentation on versioning](https://github.com/G-Node/nix/blob/master/docs/versioning.md).

### DIFF
--- a/nixio/pycore/file.py
+++ b/nixio/pycore/file.py
@@ -26,7 +26,7 @@ except ImportError:
 
 
 FILE_FORMAT = "nix"
-HDF_FF_VERSION = (1, 0, 0)
+HDF_FF_VERSION = (1, 1, 0)
 
 
 def can_write(nixfile):
--- a/nixio/pycore/file.py
+++ b/nixio/pycore/file.py
@@ -39,7 +39,7 @@ def map_file_mode(mode):
     elif mode == FileMode.Overwrite:
         return h5py.h5f.ACC_TRUNC
     else:
-        ValueError("Invalid file mode specified.")
+        raise ValueError("Invalid file mode specified.")
 
 
 def make_fapl():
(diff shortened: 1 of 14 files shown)


nixio/pycore/file.py
```diff
@@ -25,6 +25,32 @@
     CFile = None
 
 
+FILE_FORMAT = "nix"
+HDF_FF_VERSION = (1, 0, 0)
```
gicmo: Should this not be `(1, 1, 0)`?

nixio/pycore/file.py
```diff
@@ -90,16 +113,22 @@ def _open(cls, path, mode=FileMode.ReadWrite):
             path = path.encode("utf-8")
         except (UnicodeError, LookupError):
             pass
-        if not os.path.exists(path):
-            mode = FileMode.Overwrite
 
-        h5mode = map_file_mode(mode)
+        if (not os.path.exists(path) and mode == FileMode.ReadOnly):
```
gicmo: I don't think we need to outer `( )`, right?

nixio/pycore/file.py
```diff
@@ -90,16 +113,22 @@ def _open(cls, path, mode=FileMode.ReadWrite):
...
+            raise RuntimeError(
+                "Cannot open non-existent file in ReadOnly mode!"
+            )
 
-        if os.path.exists(path):
-            newfile = cls._open_existing(path, h5mode)
-        else:
+        if (not os.path.exists(path)) or (mode == FileMode.Overwrite):
```
gicmo: In line 117 you don't out `()` around the inner conditions.

nixio/pycore/file.py
achilleas-k: Yes, but not until it's changed in NIX master as well.

PR #236, opened 2016-11-15
Inverse search from Sections and Sources

achilleas-k:
Implements inverse search:
  - from metadata Sections to Blocks, Groups, DataArrays, Tags, MultiTags, and Sources.
  - from Sources to DataArrays, Tags, and MultiTags.

To accommodate this, every object now maintains a reference to its direct parent at all times.

### DIFF
--- a/nixio/pycore/section.py
+++ b/nixio/pycore/section.py
@@ -18,14 +18,13 @@ from . import exceptions
 
 class Section(NamedEntity, SectionMixin):
 
-    def __init__(self, h5group, parent=None):
-        super(Section, self).__init__(h5group)
-        self._parent = parent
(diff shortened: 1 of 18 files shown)


nixio/pycore/section.py
```diff
@@ -210,5 +209,81 @@ def parent(self):
...
+        return multi_tags
+
+    @property
+    def referring_sources(self):
+        f = self.file
+        sources = []
+        for blk in f.blocks:
+            sources.extend(src for src in blk.sources
+                           if src.metadata.id == self.id)
+        return sources
```
jgrewe: is this meant to append the generator list? in referring blocks a list is used.

Issue #186, opened 2016-07-26
Use Python enums where appropriate


achilleas-k:
DataType, LinkType, and DimensionType are currently static classes. It would be more appropriate if they were enums instead.

PR #247, opened 2017-03-02
[WIP] Changing DataType, LinkType and DimensionType from static classes to Enums

s0nskar:
Changed DataType, LinkType and DimensionType from static classes to Enums as mentioned in #186 .

### DIFF
--- a/nixio/dimension_type.py
+++ b/nixio/dimension_type.py
@@ -6,8 +6,9 @@
 # modification, are permitted under the terms of the BSD License. See
 # LICENSE file in the root of the Project.
 
+from enum import Enum
 
-class DimensionType(object):
+class DimensionType(Enum):
     Sample = "sample"
     Range = "range"
     Set = "set"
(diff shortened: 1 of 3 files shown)


nixio/dimension_type.py
```diff
@@ -6,8 +6,9 @@
 # modification, are permitted under the terms of the BSD License. See
 # LICENSE file in the root of the Project.
 
+from enum import Enum
 
-class DimensionType(object):
+class DimensionType(Enum):
     Sample = "sample"
```
s0nskar: Case should be maintained. Here `Sample="sample"` i.e in lowercase and in `link_type.py` it is in uppercase.

achilleas-k:
Unfortunately, as you've probably noticed, it's not as simple as subclassing `Enum`.
The main issue is how DataType gets converted to a C++ equivalent when being passed on to the NIX backend through the bindings. The other Enums are probably easier to handle.

If we're going to use Enums for these classes, I want it done properly and consistently. The DataType of values and arrays is going to need a lot more work to be consistent with current behaviour and the NIX backend.

achilleas-k:
Closing due to inactivity. Open new PR if necessary.

Issue #165, opened 2016-03-11
Appending objects to ProxyList gets increasingly slower


achilleas-k:
The following issue appeared while I was testing the neo-nixio writer against the sample files found at https://portal.g-node.org/neo/. I'm not sure whether I should be reporting this issue here or on the NIX backend repository instead.

The conversion of objects becomes slower as more NIX objects are created. At first I thought it was an issue with the writer, but I think narrowed it down to the `append()` operation on proxy lists, such as the `Group.multi_tags` list.

One example of this is the file https://portal.g-node.org/neo/brainwaresrc/multi_500ms_mulitrep_ch1.src, which contains roughly 1000 SpikeTrain objects (208 Segments with 5 SpikeTrains each). After a few hundred MultiTags are created, each object can take up to a couple of seconds to be converted and written.

I ran a few test scripts which simply create MultiTags on a Block and append them to a Group. The results are quite clear that as the number of objects increases, both the creation and the append operations become much slower.

PR #248, opened 2017-03-03
[Section] Disabling section parent link

achilleas-k:
Wasn't sure whether to merge this into master. I've commented out the code and added a note explaining the issue.

> This is extremely inefficient and makes writing and reading of
> files which contain a large number of objects very slow.
> In fact, it makes things increasingly slow, which amplifies
> the existing issue https://github.com/G-Node/nixpy/issues/165

The reason I say it amplifies #165 and not that it's the root cause is because the issue was opened before this change was introduced.


achilleas-k:
On second thought. Don't merge this. I'll change it so that the parent is searched for only if needed and push the changes here when done.

Keep this open.

achilleas-k:
Closing. Created issue #187. Will create new PR when the issue is resolved properly.

PR #252, opened 2017-03-23
Simplify project attributes imports in setup.py

digaru19:
Simplify the imports of the nixpy project information attributes like `AUTHOR`, `VERSION`, `CONTACT` , etc. in `setup.py` .
The implemented way is simpler than matching `regex` from the `info.py` file data.

Please review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,10 +23,12 @@ except ImportError:
 
 import sys
 import os
-import re
+
+from nixio.info import VERSION, AUTHOR, CONTACT, BRIEF, HOMEPAGE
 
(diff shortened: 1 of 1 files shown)


gicmo:
I thought there was a reason why it wasn't done like that; I have a vague memory that setup.py wasn't working like that on a system without nixpy already installed.

achilleas-k:
I thought so too and I think there was a similar case in odML, but then I discovered it was a different issue.

I'll keep it mind in case it causes any issues, but it seems to work and build without problems.

PR #254, opened 2017-04-04
PEP8 everywhere

achilleas-k:
Lots of PEP8 fixes everywhere.
All files (except `__init__.py`s) should be fully PEP8 compliant now.

Also:
- Changed the way the tests check for NIX backend availability. Instead of trying to `import nixio.core` and catching `ImportError`, simply check if NIX has the attribute (`hasattr(nix, "core")`)
- Removed star imports from tests. Now imports nixio as nix.


gicmo:
Makes some stuff less readable, but LGTM.

PR #255, opened 2017-04-06
[h5dataset] Fix for reading empty datasets

achilleas-k:
Datasets with empty shape or 0 in any dimension break h5py.
We just return the empty array when there's such data.

### DIFF
--- a/nixio/pycore/h5dataset.py
+++ b/nixio/pycore/h5dataset.py
@@ -43,7 +43,7 @@ class H5DataSet(object):
             self.dataset[:] = data
 
     def read_data(self, data, count=None, offset=None):
-        if 0 in data.shape or len(data.shape) == 0:
+        if 0 in self.dataset.shape or len(self.dataset.shape) == 0:
             return
         if count and offset:
             if sum(count) == 0 and len(data) == 0:
--- a/nixio/pycore/h5dataset.py
+++ b/nixio/pycore/h5dataset.py
@@ -43,6 +43,8 @@ class H5DataSet(object):
             self.dataset[:] = data
 
     def read_data(self, data, count=None, offset=None):
+        if 0 in data.shape or len(data.shape) == 0:
+            return
         if count and offset:
             if sum(count) == 0 and len(data) == 0:
                 return data


nixio/pycore/h5dataset.py
```diff
@@ -43,6 +43,8 @@ def write_data(self, data, count=None, offset=None):
             self.dataset[:] = data
 
     def read_data(self, data, count=None, offset=None):
+        if 0 in self.dataset.shape or len(self.dataset.shape) == 0:
+            return
```
jgrewe: Does this work? Do we need to switch the checks? In case dataset has no shape does the first check pass?
Just wondering if something like would be safer:
```
if len(self.dataset.shape) < 1 and 0 in self.dataset.shape:
         return
```

achilleas-k: I see your point, but in the worst case `shape` should be `()`, which would make `0 in shape` false (it will pass the first check with false). It's written the way it is because it started as a workaround for the weird case where the shape is `(0,)` or something like `(2, 0, 1)` (nonsensical, but valid). Both conditions (`(..., 0, ...)` and `()`) are equally rare I believe, so either order is just as valid (maybe..?).

jgrewe: in matlab something like ```(2, 0, 1)``` would be just fine, In fact it has only 2 dimensions and if you call squeeze on it, it would leave a 2D matrix. Not sure whether same things happen with numpy. I agree, from the mathematical point of view this matrix would be empty.

achilleas-k: Do you think there's a nicer way of handling these situations?
I think a more important question is whether we should disallow storing arrays with these kinds of shapes.

jgrewe: I need to correct my claim about matlab, though I am kind-of sure, I have seen this at some point. Anyway, the current state is that matlab does not have dimensions that have a zero size. Similar to numpy zeros(2,0,3) gives an empty matrix with the given shape and the size 0. 
HDF does not seem to have a problem with storing this dimension information and hdfview just opens the file. It seems, though that nix does not like it either. 

Regarding the desired behaviour in nixpy, I do not have a strong opinion. Maybe act as numpy does and return an empty variable of the given shape ```np.zeros(shape)```  instead of returning ```None```.

achilleas-k: Well, that's exactly what it does now. The `data` object is supposed to hold the data. In the standard case, the function ends with `self.dataset.read_direct(data)` and nothing is returned.
So when we do a simple `return`, the `data` array is just returned empty.

gicmo:
I personally also had swapped the conditions, but I don't think that is an issue ;)

Issue #262, opened 2017-05-07
Inconsistency between pycore and cpp bindings


delwen:
An analysis that works with the cpp bindings does not work with pycore:

```sh
│ ├─ Tags: ap1 Traceback (most recent call last):
  File "./ca-analyse.py", line 455, in <module>
    main()
  File "./ca-analyse.py", line 448, in main
    analyser.process()
  File "./ca-analyse.py", line 189, in process
    self.process_file(path)
  File "./ca-analyse.py", line 215, in process_file
    self.process_neuron(neuron)
  File "./ca-analyse.py", line 268, in process_neuron
    data = np.array([mtag.retrieve_data(pc, 0)[0] for pc in range(len(p))])
  File "/usr/local/lib/python2.7/site-packages/nixio/pycore/multi_tag.py", line 155, in retrieve_data
    offset, count = self._get_offset_and_count(ref, posidx)
  File "/usr/local/lib/python2.7/site-packages/nixio/pycore/multi_tag.py", line 112, in _get_offset_and_count
    offsets.append(self._pos_to_idx(offset[idx], unit, dim))
  File "/usr/local/lib/python2.7/site-packages/nixio/pycore/tag.py", line 145, in _pos_to_idx
    dimunit = dim.unit
AttributeError: 'SetDimension' object has no attribute 'unit'
```

delwen:
Analysis script is in [gicmo/CaAnalysis](https://github.com/gicmo/CaAnalysis).

delwen:
Applying to PR 263 and testing again now leads to:
```
  File "./ca-analyse.py", line 455, in <module>
    main()
  File "./ca-analyse.py", line 448, in main
    analyser.process()
  File "./ca-analyse.py", line 189, in process
    self.process_file(path)
  File "./ca-analyse.py", line 215, in process_file
    self.process_neuron(neuron)
  File "./ca-analyse.py", line 268, in process_neuron
    data = np.array([mtag.retrieve_data(pc, 0)[0] for pc in range(len(p))])
  File "/usr/local/lib/python2.7/site-packages/nixio-1.3.1-py2.7-macosx-10.12-x86_64.egg/nixio/pycore/multi_tag.py", line 156, in retrieve_data
    if not self._position_and_extent_in_data(ref, offset, count):
[...]

PR #263, opened 2017-05-09
FIX for Tag.retrieve_data bug

achilleas-k:
- Adds test for `Tag.retrieve_data`, ported from NIX tests.
- Adds check for `"none"` string value in Tag units.
- Checks for dimension type and doesn't attempt to access `unit` of `SetDimensions` (cause of issue #262).
- Fixes a couple of other bugs in the `retrieve_data` method that were introduced previously and went uncaught.

### DIFF
--- a/nixio/pycore/dimensions.py
+++ b/nixio/pycore/dimensions.py
@@ -5,7 +5,7 @@
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted under the terms of the BSD License. See
 # LICENSE file in the root of the Project.
-from __future__ import (absolute_import, division, print_function)
+from __future__ import (absolute_import, division)
 from numbers import Number
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


achilleas-k:
The errors that occur on Windows/Appveyor have been traced to a bug in numpy: https://github.com/numpy/numpy/issues/9216

The bug can occur when attempting to resize an array to `()`, which we do while reading data from an HDF5 dataset (see https://github.com/G-Node/nixpy/blob/master/nixio/pycore/h5dataset.py#L57).  The reason for the resize is that the resulting `data` numpy array must match the shape of the requested data. When a scalar is requested, `data` is resized to a 0d array, which is slightly different but not entirely (see https://stackoverflow.com/questions/773030/why-are-0d-arrays-in-numpy-not-considered-scalar).

The resize works as it should but crashes (sometimes) on Windows, in 64-bit interpreters, due to the aforementioned numpy bug.

PR #264, opened 2017-05-09
Add NIXPY_H5_BACKEND for runtime backend selection

gicmo:
If no backend has been set in code we can overwrite the default
backend via the NIXPY_H5_BACKEND env variable. Useful for testing
scripts/code against both backends without modifying existing
code.

NB: relies on os.getenv to return NONE for unknown variables, which
it should for both 2.7 and 3.

### DIFF
--- a/nixio/pycore/file.py
+++ b/nixio/pycore/file.py
@@ -146,6 +146,8 @@ class File(FileMixin):
         :return: nixio.File object
         """
         if backend is None:
+            backend = os.getenv("NIXPY_H5_BACKEND")
+        if backend is None:
             if CFile is None:
                 backend = "h5py"
             else:


achilleas-k:
Wouldn't `NIXPY_BACKEND` be a nicer name?

gicmo:
@achilleas-k yeah, I tried to be future proof, i.e. when there is a filesystem backend. Then this one would only control the hdf5 related backend (e.g. you set `NIXPY_BACKEND` to `"h5py"` and then you want to open a odML file or filesystem or what ...). My good our names are confusing; I wish we had chosen to cool the backend `"nix::h5"` and `"python::h5"`

gicmo:
Having said that. Maybe `NIXPY_MODULE` (or `NIXPY_BACKEND`, although that is also confusing) and let it be either "python" or "nix"?

jgrewe:
I agree with @gicmo, even though I agree that ``NIXPY_H5_BACKEND`` is not the most elegant thingy, leaving the H5 away confuses me, having read the latest comment I wonder, if I really understand what it should stand for :)

achilleas-k:
Keep it as is then.

PR #266, opened 2017-05-19
Header attribute fixes

achilleas-k:
Header attributes were defaulting to python3 types (int64 and unicode string), while NIX expects the python2 defaults (int32 and ascii string).


achilleas-k:
Closing. Separate PR coming which includes this.

PR #267, opened 2017-05-24
FIXES: Order tracking, (Multi)Tag.retrieve_data, and header format string type

achilleas-k:
Fixed the order tracking issue. The iteration order was set to `NATIVE` instead of `INC`(reasing).

This PR includes the header changes from PR #266 and fixes the issue (tested locally for combinations of python2 and python3).

It also PARTIALLY includes PR #263, but the issue is not completely resolved yet.

### DIFF
--- a/nixio/pycore/dimensions.py
+++ b/nixio/pycore/dimensions.py
@@ -5,7 +5,7 @@
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted under the terms of the BSD License. See
 # LICENSE file in the root of the Project.
-from __future__ import (absolute_import, division, print_function)
+from __future__ import (absolute_import, division)
 from numbers import Number
 
 import numpy as np
@@ -20,7 +20,6 @@ class Dimension(object):
     def __init__(self, h5group, index):
         from nixio.pycore.h5group import H5Group
         if not isinstance(h5group, H5Group):
-            print(h5group)
             raise Exception
         self._h5group = h5group
         self.dim_index = index
--- a/nixio/pycore/file.py
+++ b/nixio/pycore/file.py
@@ -9,6 +9,7 @@ from __future__ import (absolute_import, division, print_function)
 import os
 import gc
 from warnings import warn
+import numpy as np
 
 import h5py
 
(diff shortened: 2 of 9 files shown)


jgrewe:
@achilleas-k do you want to include the tag changes and the merge commit into this pull request? Would prefer to enter the danger-zone and do some rebasing :)

achilleas-k:
Sure. Keep this PR open until I rebase then.

achilleas-k:
Rebase was clean. PR updated and now includes fixes for issue #262 (it is rebased off PR #263).

achilleas-k:
_I had forgotten that this PR was rebased. I closed PR #263 and I'm copying the description of the Windows numpy issue here._

The errors that occur on Windows/Appveyor have been traced to a bug in numpy: https://github.com/numpy/numpy/issues/9216

The bug can occur when attempting to resize an array to `()`, which we do while reading data from an HDF5 dataset (see [this line](https://github.com/G-Node/nixpy/blob/2fcbad8152fdcd504172bb70d0e8baafb9b17b83/nixio/pycore/h5dataset.py#L57) in the h5dataset read function).  The reason for the resize is that the resulting `data` numpy array must match the shape of the requested data. When a scalar is requested, `data` is resized to a 0d array, which is slightly different but not entirely (see https://stackoverflow.com/questions/773030/why-are-0d-arrays-in-numpy-not-considered-scalar).

The resize works as it should but crashes (sometimes) on Windows, in 64-bit interpreters, due to the aforementioned numpy bug.

The issue now is whether we should work around the numpy bug or keep it as is and wait for the update in numpy. I'm not sure there's an elegant workaround for this and I don't really want to put in a temporary fix just for this one configuration (Py-x64 on Windows).

achilleas-k:
Once this is merged I can finalise the 1.3.1 release and then we can go ahead and do NIX and NIXPy 1.4.

jgrewe:
maybe we should add a warning to the release regarding 64bit Windows

achilleas-k:
Yeah, I'll mention it in the release "features".

PR #268, opened 2017-06-08
Test and CI changes

achilleas-k:
This PR makes a few changes to the CI and test environments

## Pytest
Switching to using pytest instead of nose by default (i.e., when running `python setup.py test`). I generally like the output formatting and configurability of pytest more.

## Travis
Updated the `--no-use-wheel` for some CI configurations to `--no-binary`, since the former is deprecated.

## Appveyor
Cleaned up the Appveyor configuration. It's no longer using miniconda and instead gets everything from pypi via pip. I also disabled (commented out) the buildings with nix since they've been failing (and allowed to fail) for too long now and it's simply a wasted of resources to let them start and fail each time. I will revisit these in the future.

It's also now possible to enable RDP for any build by adding `RDP: "yes"` to any item in the build matrix.

## Tests
Files created during tests are deleted now to keep user directories clean if and when they choose to run them.


achilleas-k:
Appveyor build failure is again attributed to the numpy bug.

achilleas-k:
Pushed a small commit for `__eq__` on the  base entity class that was generally missing.

PR #271, opened 2017-06-27
FIX Alias Range Dimension handling

achilleas-k:
This PR is primarily to fix the handling of Alias Range Dimensions in pycore but also includes some other cleanup and maintenance changes.
 
The alias dimension handling wasn't creating a link to the data array as specified by the NIX specification. Instead it was making a copy and so the two backends were behaving differently and has this small incompatibility.

Here's a nice script that creates a file with an alias dimension using both backends and then compares them using vimdiff:

```bash
#!/usr/bin/env bash

cat > createaliasdim.py << EOF
import nixio as nix
import numpy as np
import sys


writer = sys.argv[1]

# let's make up some data
event_times = np.array([1, 2, 3, 5, 8, 13, 21])

f = nix.File.open("alias_test.nix", nix.FileMode.Overwrite, writer)
b = f.create_block("test", "test")
da = b.create_data_array("the name of the data array", "nix.irregular",
                         data=event_times)
da.unit = 's'
da.label = 'this is a label'
da.append_alias_range_dimension()

f.close()
EOF

set -e

python createaliasdim.py hdf5
mv alias_test.nix alias_test_hdf5.nix

python createaliasdim.py h5py
mv alias_test.nix alias_test_h5py.nix

vimdiff <(h5dump alias_test_hdf5.nix) <(h5dump alias_test_h5py.nix)
```

NOTE: I notice there's still a lot of places where one backend will create UTF8 strings while the other does ASCII. I'm opening an issue to go through all of pycore and fix this.

NOTE2: Don't be put off by the large number of affected files. This PR also includes a small change in the naming convention for tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ matrix:
           os: linux
         - python: "3.5"
           os: linux
+        - python: "3.6"
+          os: linux
         # - language: generic
         #   os: osx
         #   before_install:
--- a/nixio/pycore/data_array.py
+++ b/nixio/pycore/data_array.py
@@ -141,8 +141,7 @@ class DataArray(EntityWithSources, DataSet, DataArrayMixin):
             raise ValueError("Cannot append additional alias dimension. "
                              "There must only be one!")
         dimgroup = self._h5group.open_group("dimensions")
-        data = self._h5group.group["data"]
-        return RangeDimension._create_new(dimgroup, 1, data)
+        return RangeDimension._create_new_alias(dimgroup, 1, self)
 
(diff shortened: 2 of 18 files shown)


achilleas-k:
Not sure what exactly broke the 2.7 `withnix` build but will have to investigate tomorrow.

achilleas-k:
Appveyor builds fail for the same reason as before: numpy bug still not fixed in latest release. I'd change it to pull and install from 'master', but that would make builds take longer and it's not a change I'd like to keep around.

Issue #273, opened 2017-08-14
Polynomial coefficients cause truncation errors for integer data arrays


achilleas-k:
The following currently fails:
```python
data = [10, 29, 33]
intarray = self.block.create_data_array("intarray", "array",
                                        nix.DataType.Int64,
                                        data=data)
intarray.polynom_coefficients = (0.0, 0.1)
np.testing.assert_almost_equal(intarray[:], np.array(data) * 0.1)
```

The data read from `intarray` gets converted to `[1, 2, 3]` because the array's dtype constrains the return value to Int64.

This doesn't happen in NIX.

PR #274, opened 2017-08-14
Fix for value truncation when applying polynomial coefficients

achilleas-k:
Fix for issue #273.

### Bug description
When reading a DataArray, a numpy array is allocated with a data type that matches the DataArray. When the DataArray has polynomial coefficients defined and the resulting values of applying the coefficients to the raw data is outside the domain of the data type, the data values are truncated.

### Example
A data array with *Integer* dtype, internal value `[11]`, and polynomial coefficients `(0, 0.1)` will return the integer value `1` instead of `1.1`.

### Fix
Always return numpy arrays of type `float64` when a DataArray has any polynomial coefficients defined.

---
I'd be interested to know if anyone thinks we could be smarter about this. For instance, should we be testing if both data and coefficients are integers and not change the type?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ pip-log.txt
 *.log
 *.kate-swp
 *.swp
+*cache
 
 # idea / pycharm files
 .idea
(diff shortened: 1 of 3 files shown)


jgrewe:
if I an not mistaken applyPolynomial in Nix does exactly the same. No smart suggestions from my side :)

gicmo:
Sounds sane to me!

PR #276, opened 2017-10-09
Removing NIX C++ bindings [dev branch]

achilleas-k:
I'm starting a new dev branch to keep track of the progress of the removal of the bindings from NIXPY and the flattening of the class hierarchy. The bindings removal started a while ago and since we all agree on going ahead with it now, I thought I'd start a dev branch where I can PR changes as I work on them, to avoid having a huge PR with all the changes in the end.

The current PR is the state of the branch as it was a few months ago. The bindings are completely gone here, but it still needs a lot of work and cleanup. There's more work that I've put into it more recently, but I think this stage is pretty good for an initial review.

It's also rebased against master. I'll regularly keep the dev branch up to date with master to make the transition easier in the end.

I know it's already quite big, so take your time with the review.


achilleas-k:
So this is huge and probably impossible to read. I can probably squash and rebase and slice it down to smaller batches build and test successfully.

I'm going to write some cross-compatibility tests starting today. Perhaps we can review those first and then base this PR and future ones off the new tests.

Issue #269, opened 2017-06-13
CompoundUnits not accepted as data unit


JuliaSprenger:
At least in the 'purest-python' branch AnalogSignals and SpikeTrains carrying pq.CompoundUnits can not be saved using the neo NixIO.

Here are two example objects:
`st = neo.SpikeTrain([1,2,3]*pq.CompoundUnit('s'), t_start=0*pq.s,
                        t_stop=10*pq.s, sampling_rate=1*pq.Hz)`
or
`anasig = neo.AnalogSignal([1,2,3]*pq.CompoundUnit('V'), t_start=0*pq.s,
                              sampling_rate=1*pq.Hz)`
can not be saved using nixio.write_block due to

>   File "/python-neo/neo/io/nixio.py", line 555, in _write_object
>     self._write_data(nixobj, attr, objpath)
>   File "/python-neo/neo/io/nixio.py", line 901, in _write_data
>     obj.unit = attr["data.units"]
>   File "/nixpy/nixio/data_array.py", line 574, in unit
>     "DataArray.unit"
> nixio.exceptions.exceptions.InvalidUnit: InvalidUnit: (V) is not SI or composite of SI units evoked at: DataArray.unit

However, keeping data arrays with their original units can be of advantage in many cases, eg for displaying or storing in an efficient way.

achilleas-k:
`V` and `s` should work, even when they're specified as "compound". I think the parentheses that pop up when a `CompoundUnit` is stringified confuses NIX.

I guess it would make sense to strip `()`, at least on the Neo-NIXIO where the quantities package is used.

JuliaSprenger:
Does this then also allow for more complex compound units like '1./30*s'?

achilleas-k:
NIX unit support is pretty limited. We don't use the quantities package or anything similar, so we rely on pattern matching which can be pretty strict at times. We restrict units to SI and derivatives, so NIX accepts anything that matches the pattern:
`{PREFIX}{UNIT}{POWER}` and combinations of that pattern, separated by either `*` or `/`.

`PREFIX` is any SI scaling prefix (e.g., `m`illi, `k`olo, `M`ega).
`UNIT` is any base unit symbol.
`POWER` is `^` followed by any sequence of digits.

This is why `(V)`is not accepted: parentheses aren't handled by any of our patterns.

Similarly, `1/30*s` wont work. In fact `1/s` doesn't work either and should be expressed in `Hz`.

[...]

PR #281, opened 2017-10-26
Allow non-SI units in Properties and DataArrays

### LINKED ISSUE #278: Support non-SI units (with limitations)
Non-SI units should be allowed. Using non-SI units in (Multi)Tags would prevent the usage of the `retrieve_data` functions, so an appropriate error should be raised.
Counterpart to G-Node/nix#704.

achilleas-k:
This PR:
- Allows any string to be set as a unit for Properties
- Allows any string to be set as a unit for DataArrays except when the DataArray has an AliasRangeDimension
    - Checks are performed both when setting a unit (checks for SI only if the DataArray has an AliasRangeDimension) and when adding an AliasRangeDimension (checks if the existing unit is SI)
- Sets a Property's or DataArray's unit to `None` if the unit is set to an empty string or a string that is only spaces.

Closes #278 and #269.

### DIFF
--- a/nixio/pycore/data_array.py
+++ b/nixio/pycore/data_array.py
@@ -141,6 +141,16 @@ class DataArray(EntityWithSources, DataSet, DataArrayMixin):
             raise ValueError("Cannot append additional alias dimension. "
                              "There must only be one!")
         dimgroup = self._h5group.open_group("dimensions")
+        # check if existing unit is SI
+        if self.unit:
+            u = self.unit
+            if not (util.units.is_si(u) or util.units.is_compound(u)):
+                raise InvalidUnit(
+                    "AliasRangeDimensions are only allowed when SI or "
+                    "composites of SI units are used. "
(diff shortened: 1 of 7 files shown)


nixio/pycore/data_array.py
```diff
@@ -242,12 +252,19 @@ def unit(self):
...
-        if u is not None:
+        if u:
             u = util.units.sanitizer(u)
+        if u == "":
+            u = None
+        util.check_attr_type(u, str)
+        if (self._dimension_count() == 1 and
+                self.dimensions[0].dimension_type == DimensionType.Range and
+                self.dimensions[0].is_alias and u is not None):
+            print("Checking if unit is SI")
```
jgrewe: you want to keep this, or is this debug code?

achilleas-k:
Woops. Yup. Leftover debug print. Removing and amending now.

PR #286, opened 2018-02-21
Read package info from json file

achilleas-k:
In the olden days, we had the `setup.py` parse the `info.py` with regular expressions to get the info.
At some point I changed this to a simple import, that worked as long as the package was installed using `pip install [...]` but failed if it was installed using `python setup.py install` and the system was missing any of the dependencies. This is because the setup essentially depended on importing the package itself and couldn't start the install process without importing the sources.

With this PR, we change both the `setup.py` and `info.py` to read the information from a JSON file that's packaged with the source files.

<details>
 <summary>Tested working using the following Dockerfile (click to expand)</summary>

```dockerfile
FROM debian:buster-slim

RUN apt update
RUN apt install -y python python3 git
RUN apt install -y python-pip python3-pip

RUN git clone --branch=setup-info https://github.com/achilleas-k/nixpy

WORKDIR nixpy

ENV inc=1

RUN python  setup.py install
RUN python3 setup.py install

RUN python  -c "import nixio as nix; print(nix.__version__)"
RUN python3 -c "import nixio as nix; print(nix.__version__)"
```
</details>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 
 env:
     global:
-        - NIX_BRANCH=master
+        - NIX_BRANCH=1.4
 
 matrix:
     include:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,3 +3,4 @@ recursive-include docs *
 graft src
 include CMakeLists.txt
 recursive-include scripts *
+include nixio/info.json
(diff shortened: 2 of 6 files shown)


jgrewe:
should we change the nix_branch variable in .travis.yml to 1.4 instead of master? In master, nix lib has been renamed to nix

achilleas-k:
THANK YOU!

I was trying to remember why the build is breaking. Changing it now.

PR #287, opened 2018-02-28
[setup] Fix minor typo in setup.py

achilleas-k:


gicmo:
Now only @stoewer is missing :D

PR #291, opened 2018-03-26
[bindings removal] Set up proof-of-concept for cross-compatibility checks

achilleas-k:
This PR removes the bindings to NIX and sets up a proof-of-concept for the cross-compatibility tests.

The general idea for the cross-compatibility tests is to create files in nixpy with a given structure and attributes and then compile and run small C++ "scripts" (so to speak) that read the file and compare against the expected structure and data. The inverse will also be implemented.

In the current PR this is done for the `name`, `type`, and `definition` attributes of `Block` and `Group`.

All `*.cpp` files under `nixio/tests/xcompat` are compiled before running any of the tests in `test_nix_compatibility.py`.

I'd like to get some feedback on this setup, the way the (C++) tests are written and built, and if you think there might be a simpler way to do this.


achilleas-k:
There's one issue with the current state of the tests: The C++ tests are compiled during import of the `test_nix_compatibility` module (I just realised). This is stupid since they still get compiled even if you don't want to run the compat tests (pytest test discovery imports the modules to determine what to run).

I'll fix this in a minute.

achilleas-k:
Done

nixio/test/test_nix_compatibility.py
```diff
@@ -0,0 +1,483 @@
+# Copyright (c) 2017, German Neuroinformatics Node (G-Node)
```
gicmo: 2018 maybe? ;)

nixio/test/test_nix_compatibility.py
```diff
@@ -0,0 +1,483 @@
...
+    """
+    run_cpp("readfile", fname)
+
+
+def run_cpp(command, *args):
+    cmdbin = os.path.join(BINDIR, command)
+    cmdargs = [cmdbin]
+    cmdargs.extend(args)
+    proc = Popen(cmdargs, stdout=PIPE, stderr=PIPE,
+                 env={"LD_LIBRARY_PATH": "/usr/local/lib"})
```
gicmo: Does that work on Windows? Can we avoid it by using the rpath linker flag?

PR #293, opened 2018-04-26
More crosscompatibility tests: Full feature test

achilleas-k:
The main part of this PR is a new comprehensive compatibility test. On the Python side, the `test_full_file` function creates a file using every NIX feature. The corresponding `readfullfile.cpp` reads in the file and verifies the structure and data is as expected.

This is one sided: NIXPy writes, NIX reads. The inverse test will come in a later PR (likely the next one).

It would be nice if you could read the Python writer function carefully and let me know if there are any features (object properties, relationships, etc) that aren't being tested, so that I can include them in the test. I want this test to be as fully featured as possible.

Smaller tests for individual features will also be added in future PRs.


nixio/test/test_nix_compatibility.py
```diff
@@ -435,48 +450,171 @@ def test_full_write(tmpdir):
...
     nixfilepath = os.path.join(str(tmpdir), "filetest.nix")
     nix_file = nix.File.open(nixfilepath, mode=nix.FileMode.Overwrite,
                              backend="h5py")
 
     block = nix_file.create_block("blockyblock", "ablocktype of thing")
-    block.description = "I am a test block"
-
-    group = block.create_group("group", "group type of thing")
-    group.description = "I am a test group"
+    block.definition = "I am a test block"
```
jgrewe: haha, very good catch! kind of embarrassing that we did not notice before

nixio/test/test_nix_compatibility.py
```diff
@@ -435,48 +450,171 @@ def test_full_write(tmpdir):
...
+    tag.extent = [1920, 1080]
+    tag.units = ["mm", "mm"]
+
+    png = block.create_data_array("some-sort-of-image?", "png",
+                                  shape=(3840, 2160))
+    tag.create_feature(png, nix.LinkType.Indexed)
+
+    newmtpositions = block.create_data_array("nu-pos", "multi-tag-positions",
+                                             shape=(10, 3),
+                                             dtype=nix.DataType.Bool)
```
jgrewe: this should rather be double instead of bool

Issue #296, opened 2018-05-09
Error while reading from specific dimensions of DataArray


adityachivu:
I am storing a 4D matrix of type np.uint16 in a DataArray. When I try to read a subset of the data, I get the following error in some cases:

```
Traceback (most recent call last):
  File "/path/to/env/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2910, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-39-bc26bec09c5a>", line 23, in <module>
    img = data[0, :, 0, 0] # OSError
  File "/path/to/env/lib/python3.5/site-packages/nixio/data_array.py", line 137, in __getitem__
    self._read_data(raw, count, offset)
  File "/path/to/env/lib/python3.5/site-packages/nixio/pycore/data_array.py", line 82, in _read_data
    super(DataArray, self)._read_data(data, count, offset)
  File "/path/to/env/lib/python3.5/site-packages/nixio/pycore/data_array.py", line 28, in _read_data
    dataset.read_data(data, count, offset)
  File "/path/to/env/lib/python3.5/site-packages/nixio/pycore/h5dataset.py", line 58, in read_data
    self.dataset.read_direct(data, sl)
  File "/path/to/env/lib/python3.5/site-packages/h5py/_hl/dataset.py", line 657, in read_direct
    self.id.read(mspace, fspace, dest, dxpl=self._dxpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 181, in h5py.h5d.DatasetID.read
  File "h5py/_proxy.pyx", line 130, in h5py._proxy.dset_rw
  File "h5py/_proxy.pyx", line 84, in h5py._proxy.H5PY_H5Dread
OSError: Can't read data (src and dest data spaces have different sizes)
```

The following is an example for a NIX file created with DataArray of size `[5, 10, 15, 20]` The same Errors occur:

```python
import nixio as nix
import numpy as np
import os

file_name = 'nix_test.hd'

if not os.path.isfile(file_name):

    nf = nix.File.open(file_name, nix.FileMode.Overwrite)
    block = nf.create_block('test', 'test')
    data_arr = block.create_data_array('test', 'test', 
                                        data=np.random.randint(65000, size=[5, 10, 15, 20]))
[...]

PR #297, opened 2018-05-14
Fix for data indexing and slicing

achilleas-k:
A relatively simple fix for the multidimension indexing/slicing issue (see #296).

The problem was in the dimensionality "fixing" that I had written for the cases where the number of dimensions to be returned were fewer than the number of dimensions of the underlying data. The old fix relied on padding dimensions and made incorrect assumptions about why the data dimensions didn't match.

Now we use the count tuple to set the shape of the data, since count always matches the shape of the data to be read. The returned data is then reshaped prior to returning from the function to fit the requested data shape.

I added a test, copied from @adityachivu's example in the issue.

### DIFF
--- a/nixio/pycore/h5dataset.py
+++ b/nixio/pycore/h5dataset.py
@@ -50,11 +50,10 @@ class H5DataSet(object):
                 return data
             datashape = data.shape
             sl = util.co_to_slice(count, offset)
-            if (isinstance(sl, tuple) and
-                    np.ndim(data) != len(sl) and
-                    count[-1] == 1):
-                    # data needs an extra dimension of size 1 to match slice
-                    data.resize(datashape + (1,))
+            if isinstance(sl, tuple) and np.ndim(data) < len(sl):
+                # data shape needs dimension padding to match dataset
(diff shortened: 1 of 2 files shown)


achilleas-k:
Making some changes to the Travis configuration. Don't want to keep testing the bindings on 3.4 and Travis doesn't have boost-python for anything > 3.4.

nixio/pycore/h5dataset.py
```diff
@@ -50,11 +50,10 @@ def read_data(self, data, count=None, offset=None):
...
             datashape = data.shape
             sl = util.co_to_slice(count, offset)
-            if (isinstance(sl, tuple) and
-                    np.ndim(data) != len(sl) and
-                    count[-1] == 1):
-                    # data needs an extra dimension of size 1 to match slice
-                    data.resize(datashape + (1,))
+            if isinstance(sl, tuple) and np.ndim(data) < len(sl):
+                # data shape needs dimension padding to match dataset
+                # diimensionality
```
jgrewe: ha, found something :) "diimensionality"

PR #304, opened 2018-06-18
Slice data by index or dimensional data

### LINKED ISSUE #298: Feature Request: Retrieve data with custom extent
Hey guys,
I think it would be handy if you could retrieve data of a tag (or multi tag) from position to a custom extent.
Sometimes the interval between position and extent is too short (e.g. when using short stimuli). With some help from Jan I managed to do use using some extra code. But it would be nice to have this as a included function.
Best
Nils

achilleas-k:
This PR brings the ability to retrieve a data slice from a DataArray based on either indexes (like regular slicing) or based on the dimensional data of the array. The method is currently called `get_slice()` but I'm open to suggestions for renaming it. I'd also rename the accompanying enum, `DataSliceMode`, to match whatever we end up calling the method, if we do rename it.

There's a small test that's based on [my comment](https://github.com/G-Node/nixpy/issues/298#issuecomment-397801984) from the issue where we talked about this feature. Have a look at that and let me know if I missed anything.

Also, as discussed, the method returns a DataView. This required splitting out the definition of the DataSet from the data_array file.

A small fix is also included in this PR: The dimension.index_of() methods would return floats sometimes. They now both return integers explicitly.

Closes #298

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,7 @@ environment:
   matrix:
     # Python 2.7 32 bit
     - PYTHON: "C:\\Python27"
+      pyver: 2
       bits: 32
       withnix: "--without-nix"
     # - PYTHON: "C:\\Python27"
@@ -21,6 +22,7 @@ environment:
 
     # Python 3.6 32 bit
     - PYTHON: "C:\\Python36"
+      pyver: 3
       bits: 32
       withnix: "--without-nix"
     # - PYTHON: "C:\\Python36"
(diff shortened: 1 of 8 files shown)


achilleas-k:
So we good on the naming?

jgrewe:
I have no better idea ;)

one thing that crossed my mind is that is has the same performance issues as the the original retrieve data functions... That is, if many slices are requested with a rangeDimension, it will always need to get the ticks to find the indices.

achilleas-k:
Ah yeah, forgot all about that.

Shouldn't be *too hard* (_famous last words_) to vectorise the `index_of()` functions and then the `get_slice()`. Should I add to this PR or open a new one?

jgrewe:
hehe, a separate one? Could be solved in one go with the retrieve data changes (return of multiple slices)

PR #306, opened 2018-07-05
CI and test cleanup

achilleas-k:
The plan was for this PR to enable cross-compatibility tests on macOS (Travis) and Windows (Appveyor).
This PR does not enable cross-compatibility tests on macOS and Windows.

I wont go into the issues here. I'll write up a couple of issues tomorrow.

This PR is mostly about cleaning out CI configurations and making them a bit simpler and straightforward.


nixio/test/test_nix_compatibility.py
```diff
@@ -20,8 +20,12 @@
...
 
 # skip these tests if nix isn't available
-pytestmark = pytest.mark.skipif("skip()",
-                                reason="Compatibility tests require NIX")
+if pytest.config.getoption("--force-compat"):
+    print("Forcing compatibility tests")
+    maketests(BINDIR)
+else:
+    pytestmark = pytest.mark.skipif("skip()",
+                                    reason="Compatibility tests require NIX")
```
jgrewe: tiny thing that crossed my mind: is the message self-explanatory? Maybe something like " require nix C++ library"?

achilleas-k: Sure, I could change it, though it's not really important. The compatibility tests aren't meant for anyone but us. I'll change it anyway, though.

PR #308, opened 2018-07-09
Remove mixins

achilleas-k:
Things are starting to clear up from the removal of the bindings now:

- **Mixin classes removed:** The mixin classes were originally used to augment the bindings and then became superclasses for the pycore objects. The mixin methods have no been moved into the pycore classes and the runtime injections have been removed

- **Pycore submodule removed:** All NIX classes in pycore have been moved to the root of the module.

- **New submodule HDF5:** The two classes that wrap the HDF5 objects (H5Group and H5DataSet) are no part of a module called HDF5.

- **Top level import:** Since we no longer have this runtime-determined class type situation (i.e., core.Block vs pycore.Block), all NIX objects are available at the top level module import (e.g., nix.Block, nix.Group). This makes it easier to do type checks.


achilleas-k:
Ah, good catch. That was changed because, after merging the mixins into the main classes, I copied the files from `pycore` into the root and they overwrote the originals which had the old 2014 copyright notice. I'm not entirely sure what the copyright date should reflect exactly. Should I revert it?

PR #309, opened 2018-07-11
add a metadata printing functionality

jgrewe:
which is one of the features people here are missing because it is so hard to browse the tree manually. I just punched it into the pycore files and need some ideas, feedback

So far it uses print but one could probably  think of more elegant ways.

Don't be shy!

### DIFF
--- a/nixio/pycore/property.py
+++ b/nixio/pycore/property.py
@@ -148,8 +148,7 @@ class Property(Entity, PropertyMixin):
     def __values_to_string(self, max_length=80):
         unit = "" if self.unit is None else self.unit
         value_str = ""
-        for i, v in enumerate(self.values):
-            value_str += ("%s%s") % (", " if i > 0 else "", v.to_string(unit=unit))
+        value_str = ", ".join(v.to_string(unit=unit) for v in self.values)
 
         if len(value_str) <= max_length - 4:
             value_str = "[ %s ]" % (value_str)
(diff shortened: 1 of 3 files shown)


nixio/pycore/property.py
```diff
@@ -144,3 +144,22 @@ def __str__(self):
 
     def __repr__(self):
         return self.__str__()
+
+    def __values_to_string(self, max_length=80):
+        unit = "" if self.unit is None else self.unit
+        value_str = ""
+        for i, v in enumerate(self.values):
+            value_str += ("%s%s") % (", " if i > 0 else "", v.to_string(unit=unit))
```
achilleas-k: For this bit, instead of figuring out where to add a leading comma (by checking `i > 0`), it would be simpler to make a list of value strings and then just do `", ".join(vstr_list)`.

The whole thing can be: `value_str = ", ".join(v.to_string(unit=unit) for v in self.values)`

nixio/pycore/section.py
```diff
@@ -302,3 +302,20 @@ def referring_sources(self):
...
+        for p in self.props:
+            p.pprint(current_depth, indent, max_length=max_length)
+        if max_depth == -1 or current_depth < max_depth:
+            for s in self.sections:
+                s.pprint(current_depth+1, max_depth=max_depth, indent=indent, max_length=max_length)
+        elif max_depth == current_depth:
+            child_sec_indent = spaces+" "*indent
+            more_indent =  spaces+" "*(current_depth+2*indent)
+            for s in self.sections:
+                print ("%s  %s [%s] --- %s\n%s[...]" % (child_sec_indent, s.name, s.type,
```
achilleas-k: Remove the in `print (`

PR #313, opened 2018-07-16
odML upgrade to v1.4

mpsonntag:
This PR introduces odML at format version 1.4 to nixpy.

- The `create_section` and `create_property` methods now feature an optional `id` parameter. Valid UUIDs can now be assigned to a new `Section` or `Property` to enable import of existing odML data.
- The `value.py` file has been renamed to `datatype.py`, all Imports of `DataType` have been adjusted accordingly.
- The `Value` class has been removed.
- The `Value` properties `uncertainty` and `reference` have been moved to `Property`.
- New properties `dependency`, `dependency_value`, `value_origin` and `odml_type` have been added to `Property`.
- `Property.value` has been rewritten and now always returns a tuple of datatype-homogeneous values. The values of a `Property` can be changed, rewritten and reset, but the datatype e.g. `float` or `boolean` of a `Property.value` cannot be changed.
- `Property.odml_type` supports setting odml types as they are used in the [standalone implementation of odml](https://github.com/G-Node/python-odml). Currently  `int`, `float`, `boolean`, `string`, `text`, `url`, `person`, `time`, `date` and `datetime` are supported. An `odml_type` can only be set, if it can be properly mapped to the `Property.value.h5dataset.dtype`. `text`, `url`, `person`, `time`, `date` and `datetime` are interpreted as `DataType.String`.
- `reference` has been added to `Section`.
- `mapping` has been removed from `Section` and `Property`.
- Affected existing tests have been adjusted, basic tests for new properties have been added.


achilleas-k:
Should open an issue in NIX to add the `odml_type` attribute to Properties there too.

nixio/section.py
```diff
@@ -75,39 +79,120 @@ def create_section(self, name, type_):
...
+            for v in vals:
+                if DataType.get_dtype(v) != dtype:
+                    raise exceptions.InvalidAttrType(single_val, DataType.get_dtype(v))
+
+        prop = Property._create_new(self, properties, name, dtype, oid)
+        prop.values = vals
+
         return prop
 
+    def has_property_by_name(self, name):
```
achilleas-k: These methods (`has_property_by_name`, `get_property_by_name`, etc) are the old-style ProxyList methods and aren't used anymore. This functionality is now handled by the Container class which lives on the Section as `_properties` and is exposed through `props`.

nixio/property.py
```diff
@@ -7,29 +7,105 @@
...
+    datetime = 'datetime'
+    date = 'date'
+    time = 'time'
+
+    def __str__(self):
+        return self.name
+
+    def compatible(self, value):
+        """
+        compatible returns True or False dependent on whether a
```
jgrewe: depending on...

nixio/property.py
```diff
@@ -7,29 +7,105 @@
...
+        passed in value can be mapped to an OdmlType or not.
```
jgrewe: without "in"

nixio/property.py
```diff
@@ -45,95 +121,168 @@ def definition(self, d):
...
+
+    @property
+    def uncertainty(self):
+        return self._h5dataset.get_attr("uncertainty")
+
+    @uncertainty.setter
+    def uncertainty(self, uncertainty):
+        # Use int as type check for now but check whether
+        # it should be changed to float.
+        util.check_attr_type(uncertainty, int)
```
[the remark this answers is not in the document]
jgrewe: good catch! yes, should accept floating point numbers

PR #318, opened 2018-07-18
Minor metadata fixes

achilleas-k:
Some small fixes and a bit of cleanup for the new metadata property stuff.


achilleas-k:
I've upped the file format version to `1.2.0`. I don't know if this is the right number right now. In NIX, the version was bumped to 1.1.1 when Christian was writing the DataFrame. I wanted a format version that signifies a breaking change (metadata are not backwards compatible now). If we make the library able to read old-style metadata, we could make this 1.1.2, I guess?

For now, I just want files written using the dev branch to break on old versions of NIXPy. What do you think?

gicmo:
If I read [versioning](https://github.com/G-Node/nix/blob/master/docs/versioning.md) correctly and you are saying that a new library can't read an old file anymore, I think it might need to be `2.0.0`, no?

achilleas-k:
... oh yeah

achilleas-k:
well... now the nix compatibility tests fail because of the file format version...

achilleas-k:
We merging this or wat?

PR #320, opened 2018-07-25
Delete all references when deleting objects from primary Container

achilleas-k:
When objects can be referenced from multiple locations (e.g., a LinkContainer in a Group, a Source in a data object's `sources`, an object's `metadata`, etc), removing them requires scanning through the tree for references to the same object and deleting them. This PR adds a `delete_all` method to the `H5Group` backend class and uses it through the `__delitem__` methods in the Container class and its derivatives. The `delete_all` method scans through the HDF5 file and removes all groups which match a given set of entity IDs.

The `__delitem__` methods have slight variations depending on the type of Container:
- In the base **Container** class, it simply scans the object tree, starting from the parent Block, and deletes all references to the object being deleted.
- In the **SourceContainer** class, it uses `find_sources` to collect the IDs of all children of the Source that's being deleted, and sends those IDs to the `delete_all` method in the backend, starting from the parent Block.
- Similarly, in the **SectionContainer** class, it uses `find_sections` to collect the IDs of all children of the Section that's being deleted, and sends those IDs to the `delete_all` method in the backend, starting from the File root.

_Tests with all the weird cases are included_

### DIFF
--- a/nixio/container.py
+++ b/nixio/container.py
@@ -101,7 +101,7 @@ class Container(object):
 class SectionContainer(Container):
     """
     SectionContainer extends Container with a new __delitem__ method.
-    When a Section is deleted, all child sources need to be deleted
+    When a Section is deleted, all child sections need to be deleted
     individually to make sure all their references are removed.
     """
     def __delitem__(self, item):
--- a/nixio/container.py
+++ b/nixio/container.py
@@ -154,10 +154,19 @@ class LinkContainer(Container):
 
     def __contains__(self, item):
         # need to redefine because of id indexing/linking
-        if isinstance(item, self._itemclass):
-            return item.id in self._backend
+        if hasattr(item, "id"):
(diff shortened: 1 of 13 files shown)


achilleas-k:
With the new tests, I think we cover all cases where objects can be linked/referenced from their non-primary location.

nixio/container.py
```diff
@@ -83,6 +98,55 @@ def items(self):
             yield item.id, item
 
 
+class SectionContainer(Container):
+    """
+    SectionContainer extends Container with a new __delitem__ method.
+    When a Section is deleted, all child sources need to be deleted
+    individually to make sure all their references are removed.
```
jgrewe: probably child *sections* ;)

achilleas-k: Well fine, if you want to be all _correct_ about it

PR #321, opened 2018-08-08
Bugfix: Set dimension.index properly on __init__

achilleas-k:
DimensionContainer failed to set the dimension properly when
initialising a Dimension that was being loaded from the file. It was
looking for an "index" attribute and was returning None. Should be using
the name of the HDF5 Group object.

Added tests to catch regressions.

Wrapping index in int() when returning and when setting the object
attribute to catch unwanted behaviour.

### DIFF
--- a/nixio/dimensions.py
+++ b/nixio/dimensions.py
@@ -28,7 +28,7 @@ class DimensionContainer(Container):
             DimensionType.Sample: SampledDimension,
             DimensionType.Set: SetDimension,
         }[item.get_attr("dimension_type")]
-        idx = item.get_attr("index")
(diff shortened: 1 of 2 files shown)


jgrewe:
keeps me wondering whether the name should really start at 1...

achilleas-k:
I know what you mean. I wonder how many things would break if we changed it. Probably nothing in the lib itself, but perhaps some scripts? Though dimension retrieval on the user side is still 0-based, so that's probably fine too. I guess the only issue is that it's a format change and dimensions would mismatch if a new lib would open an old file.

PR #337, opened 2018-10-11
Fix for retrieve_data when MultiTag positions array is 1D

achilleas-k:
1D position arrays are handled as a special case. This used to work
before because the DataArray would allow indexing with more indexes than
the length of the array itself. The new DataArray indexing function
doesn't allow this and tries to follow numpy array indexing rules.

Test added to catch regression.

### DIFF
--- a/nixio/multi_tag.py
+++ b/nixio/multi_tag.py
@@ -138,7 +138,10 @@ class MultiTag(BaseTag):
                 ext_size[1] > len(data.dimensions)):
             raise incdim_exception
 
-        dimpos = positions[index, 0:len(data.dimensions)]
+        if len(pos_size) == 1:
+            dimpos = positions[0:len(data.dimensions)]
+        else:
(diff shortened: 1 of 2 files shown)


achilleas-k:
Here's a script demonstrating how the issue arose:
```python
import h5py
import nixio as nix
import numpy as np


# one-dimensional array
array = np.array([10, 20, 30])
print(array, array[0])

# one-dimensional NIX DataArray
with nix.File.open("/tmp/deleteme.nix", mode=nix.FileMode.Overwrite) as nfile:
    blk = nfile.create_block("B", "B")
    da = blk.create_data_array("da", "da", data=array)
    print(da, da[0])
    print(da, da[0, 0, 0, 0])  # this works in master

# one-dimensional H5 dataset
with h5py.File("/tmp/deleteme.h5", mode="w") as hfile:
    ds = hfile.create_dataset(name="ONED", data=array)
    print(ds, ds[0])
```

PR #339, opened 2018-10-24
tiny fixes to SampledDimenision util functions

jgrewe:
* fix axis function to work properly if count and start index are given
* ensure that index_of returns an int. previously this ``` array[dim.index_of(1.0) : dim.index_of(2.0)]``` did not work without  casting to int


achilleas-k:
It may be time to ignore bindings errors and just move on to the new version without bindings.

achilleas-k:
Awesome. Merging.

PR #351, opened 2018-11-09
[DataFrame] write_to_csv and write_to_pandas

hkchekc:
add two functions to export Nixpy dataframes to panda dataframe and csv file

### DIFF
--- a/nixio/data_frame.py
+++ b/nixio/data_frame.py
@@ -154,6 +154,7 @@ class DataFrame(Entity, DataSet):
         count = len(self)
         return count
 
+<<<<<<< 9d6b672635a243969a54fdd70cbde98da7eecd14
     def write_to_csv(self, filename, mode='w'):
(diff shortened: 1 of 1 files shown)


jgrewe:
@hkchekc as I see it the tests fail because pandas is not installed on the test machines. Shouldn't it suffice to add it as a dependency to the setup.py?

achilleas-k:
I think something like that (exporting to Pandas) should be part of the planned external library of nix-related tools (plotting, etc). I wouldn't want to add Pandas as a dependency to NIXPy, not even as an optional.

I already talked about this with Chek which is the reason for the final commit removing the Pandas-related functionality. I thought I'd mention it here as well to document the decision.

jgrewe:
true, fair enough. Just created a repo for that purpose :)
https://github.com/G-Node/nixworks

PR #354, opened 2018-11-29
support create with list type data

hkchekc:
support create with list type data,

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -12,6 +12,7 @@ try:
 except ImportError:
     from sys import maxsize as maxint
 import numpy as np
+from six import string_types
 
 from .util import find as finders
 from .compression import Compression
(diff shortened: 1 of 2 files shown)


nixio/block.py
```diff
@@ -197,7 +197,7 @@ def create_data_frame(self, name, type_, col_dict=None, col_names=None,
                                   " with col_dict or (col_names+col_dtypes)")
             col_dict = dict((str(nam), dt) for nam, dt in zip(col_names, col_dtypes))
         data_frames = self._h5group.open_group("data_frames")
-        shape = data.shape
+        shape = len(data)
```
jgrewe: I wonder, if we need more checking here.  if ``data`` data is a numpy matrix then len might not be the measure you want. If data must be 1d, this should be checked.
Something else that  puzzled me, ``data`` has a default ``None`` which will lead to ``ValueError``. If data is not optional i would prefer to not have it as a named argument with default.

achilleas-k: I suggested `len()` because the data in a DataFrame should be one-dimensional, where each value is a record (a struct, a row of the table). However, checking if it's two dimensional is tricky. If a user supplies a list of lists, that should be valid, assuming the internal lists fit the column types. List of tuples should also be acceptable. If all the columns have the same datatype, a 2D array might also work. Regardless, the length of the data (if any are provided) should be used for the initial size of the underlying dataset. Any second dimension will be converted to the column data.

As for the None case, data should be optional eventually. It should be possible to create an empty DataFrame the same way it's possible to create an empty DataArray and append values to it.

jgrewe: fair enough :+1:

PR #359, opened 2019-01-09
fixed tutorial pages source

JiriVanek:
new tutorial for nix no bindings dev. need to build it with sphing and push it to gh-pages

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -20,8 +20,7 @@ The following sections should help you getting started with nixpy.
 
 API Documentation
 =================
-
-TODO write something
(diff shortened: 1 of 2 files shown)


achilleas-k:
Not sure we want to tell people to install the prerelease in the release instructions. I know this is in the dev branch, but we would have to revert the change once we move the dev (1.5) stuff to master and it gets a proper release.

I think a better idea would be to add a line that the **beta** release of 1.5 can be installed with the `--pre` flag, rather than having it as the standard way to install. We could keep it there permanently as an instruction for people who want to try beta or alpha releases.

PR #360, opened 2019-01-17
Codestyle cleanup and compatibility test selection changes

achilleas-k:
### Run compatibility tests only when requested

The compatibility tests are no longer run by default when NIX is detected. They can be run by specifying --nix-compat. If specified, the tests are compiled and any errors will cause failure of the
compatibility tests. If the option is not specified, the tests are skipped, even if NIX is available.

Compatibility tests were always compiled when NIX was available, even when they are unselected, and compilation would slow down the tests. With this change, they are only compiled and run when explicitly enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,9 +77,9 @@ install:
 
 script:
   - if [[ "${coveralls}" == 1 ]]; then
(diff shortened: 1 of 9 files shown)


achilleas-k:
Merged into master via #361

PR #378, opened 2019-03-15
Docstrings for DataFrame and Validator

hkchekc:
Add the descriptions  to the DataFrame and Validator methods and properties. 

Add the link to metadata for DataFrame 

Delete the __eq__ and __hash__  for some objects and their parent Class entity already have the exactly same method.

### DIFF
--- a/nixio/data_frame.py
+++ b/nixio/data_frame.py
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
-
+# Copyright © 2019, German Neuroinformatics Node (G-Node)
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted under the terms of the BSD License. See
+# LICENSE file in the root of the Project.
 from __future__ import (absolute_import, division, print_function)
 try:
     from collections.abc import Iterable
--- a/nixio/data_frame.py
+++ b/nixio/data_frame.py
@@ -79,8 +79,7 @@ class DataFrame(Entity, DataSet):
 
     def append_rows(self, data):
         """
-        Append a new row to the DataFrame
-        In Python2, the data supplied must be iterable (not np arrays)
+        Append a new row to the DataFrame. The data supplied must be iterable.
 
         :param data: The new row
         :type data: array-like data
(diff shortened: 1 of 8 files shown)


nixio/validate.py
```diff
@@ -67,7 +78,12 @@ def form_dict(self):
                     tag['features'].append(fea_dict)
 
     def check_file(self):
+        """
+        Check if the file meet the nix requirements on file level.
```
achilleas-k:
```suggestion
        Check if the file meets the NIX requirements at the file level.
```

Same for the rest.

nixio/validate.py
```diff
@@ -80,13 +96,25 @@ def check_file(self):
         return self.errors
 
     def check_blocks(self, block, blk_idx):
+        """
+        Check if the file meet the nix requirements on block level.
```
achilleas-k:
```suggestion
        Check if the file meets the NIX requirements at the block level.
```

nixio/data_frame.py
```diff
@@ -248,3 +249,31 @@ def df_shape(self):
         df_shape = tuple(df_shape)
         self._h5group.set_attr("df_shape", df_shape)
         return self._h5group.get_attr("df_shape")
+
+    # metadata
+    @property
+    def metadata(self):
+        """
+
```
achilleas-k: Remove empty leading line in docstring

nixio/data_frame.py
```diff
@@ -68,7 +78,13 @@ def append_column(self, column, name, datatype=None):
         self.write_direct(farr)
 
     def append_rows(self, data):
-        # In Python2, the data supplied must be iterable (not np arrays)
+        """
+        Append a new row to the DataFrame
+        In Python2, the data supplied must be iterable (not np arrays)
```
achilleas-k: What happens in Python 2 with np.arrays?

hkchekc: Could not find any problems with Python2. Will delete the sentence.

On the other hand, I am thinking if the DataFrame.dtype property should return the DataFrame._h5group.group["data"].dtype (np.dtype) instead of the raw list of dtypes now. 
As it is more helpful to create np arrays with same compound data type before writing and appending new rows.

hkchekc: I guess the original problem I had in mind is that the np arrays in python 2 have less of dynamic typing and probably there was some problem when I previously test it. But if the array is created with compound type, there is no problem

achilleas-k:
Looks good thanks.  Merging!

Issue #367, opened 2019-01-24
Copy objects between files


achilleas-k:
Add the ability to create NIX objects in one file from objects in another file. This would allow copying of objects between files. The copy should optionally carry over the UUID of the object as well (clone the object).

The operation should be performed recursively, so copying a `Block` would also copy an entire hierarchy of objects.

This functionality will also be used to perform version upgrades/conversions between file format revisions.

achilleas-k:
Implemented in PR #379.

PR #379, opened 2019-03-20
Copy function for all entities

hkchekc:
Refer to the Issue #367 

This PR adds a function on the entity level to allow copying NIX objects between files. This copy is recursive and the data in the objects is also carried over.

The UUID will always be carried over.

Definitely need some tests.

How should one deal with the links to the group in the block is opened to discussion. (I have never tried just copying a non-empty group without copying the contents in the block)

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -74,7 +74,8 @@ class Block(Entity):
         if copy_from:
             if not isinstance(copy_from, MultiTag):
                 raise TypeError("Object to be copied is not a MultiTag")
-            id = self._copy_objects(copy_from, "multi_tags", keep_copy_id)
+            id = self._copy_objects(copy_from, "multi_tags",
+                                    keep_copy_id, name)
             return self.multi_tags[id]
 
         util.check_entity_name_and_type(name, type_)
@@ -106,7 +107,7 @@ class Block(Entity):
         if copy_from:
             if not isinstance(copy_from, Tag):
                 raise TypeError("Object to be copied is not a Tag")
-            id = self._copy_objects(copy_from, "tags", keep_copy_id)
+            id = self._copy_objects(copy_from, "tags", keep_copy_id, name)
             return self.tags[id]
 
         util.check_entity_name_and_type(name, type_)
@@ -185,7 +186,8 @@ class Block(Entity):
         if copy_from:
             if not isinstance(copy_from, DataArray):
                 raise TypeError("Object to be copied is not a DataArray")
-            id = self._copy_objects(copy_from, "data_arrays", keep_copy_id)
+            id = self._copy_objects(copy_from, "data_arrays",
+                                    keep_copy_id, name)
             return self.data_arrays[id]
 
         if data is None:
@@ -246,7 +248,8 @@ class Block(Entity):
         if copy_from:
             if not isinstance(copy_from, DataFrame):
                 raise TypeError("Object to be copied is not a DataFrame")
-            id = self._copy_objects(copy_from, "data_frames", keep_copy_id)
+            id = self._copy_objects(copy_from, "data_frames",
+                                    keep_copy_id, name)
             return self.data_frames[id]
 
         util.check_entity_name_and_type(name, type_)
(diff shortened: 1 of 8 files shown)


achilleas-k:
Nice initial proof of concept.  Seems to work for valid cases.  Needs more sanity checks though.  For instance, it allows attaching a DataArray to another DataArray.

I'm also not sure about the API.  I'm thinking it might make more sense to have analogues to the `append` methods for Group containers.  For instance

```python
# copy block from one file to another
block = orig.blocks[0]
dstfile.append_copy(block_from_other_file)  # could also be add_copy()

# similarly, to copy a DataArray from one file to another
blk.data_arrays.append_copy(da_from_other_file)
```

This way we could check the type of object coming in and where it's being added.

@jgrewe & @gicmo thoughts?

hkchekc:
@achilleas-k It is make sense to make it a append_copy function. I prefer to stay on the entity level instead of the container level because after all this is somehow creating "new" object instead of making links. so the append_copy can look like
blk.append_copy(da_from_somewhere_else)

achilleas-k:
I get your point about having ti be an analogue to the `create_x` functions.  I'm not sure about having the method defined on the `Entity`.  Most objects are created on the `Block` and in the current function there's a lot of logic going into figuring out what kind of object we are copying.  `Entity` is the base class for most objects in NIX but we only attach objects onto 2 of those (`Block` and `Section`) as well as 1 that's not an entity (`File`).  I'm ignoring `Dimension`s and `Feature`s for now because I'm not sure we want to copy those on their own.  Parsing the `type()` of an object as a string is also a bad idea.

If we're treating the copy function like the `create_x` functions (which I agree we should), I think they should be either alongside them (so, `copy_data_array()` would be in the `Block` class) or maybe even just overloaded versions of the create methods (e.g., `block.create_data_array(copyfrom=other_da)`).

hkchekc:
Not sure what is a better way.

Should we add an copy_object method on block.py or separate all those copy_data_array / copy_tag methods?

hkchekc:
Some changes:

1. The copy function is not any more in the Entity file. I divide it to copy_x functions in Files, Blocks and Sections

2. Support the copy of sections and properties (nested)

3. Give option to users to change the UUID while copying the object. The UUID changes are also recursive.


__

achilleas-k:
For consistency with the odML clone function, I think instead of `change_id=False` we could have `keep_id=True` in the args of the copy functions.

See https://github.com/G-Node/python-odml/blob/c7a63526785fb5149b9ae01d66e96d214dd13976/odml/section.py#L403

jgrewe:
Hi @hkchekc, thanks for your work! Have a few thoughts about it:
* I like it better to have the copy_xxx methods not on entity.
* I like @achilleas-k idea with the overloaded version. It keeps the API a bit cleaner 
* Should we add a `recursive = True` flag to the section copy to indicated whether it should copy recursive?
*  quite a bit of code repetition, can this be avoided?

hkchekc:
@jgrewe @achilleas-k 

As @jgrewe suggested I add an recursive flag sections.I also group up the repetitions a bit. Also, change_id is changed to keep_id

The major thing is that I make the copy function overload on the create_x function. It is on one hand keeping the API cleaner. On the other hand, it looks quite confusing comparing to separated functions at least for me.

One more question is that is copying objects to the same parent make sense? Do we have to support for example copying blk1.data_array[0] to blk1?

hkchekc:
@jgrewe @achilleas-k

As @jgrewe suggested I add an recursive flag sections. Also, change_id is changed to keep_id. 

The major thing is that I make the copy function overload on the create_x function. It is on one hand keeping the API cleaner. On the other hand, it looks sort of confusing, not sure which is better. 

I make the name and type_  and some other compul. parameters positional without additional checks because if it is not a copying action, an error will be throw later by _create_new function will throw an error.

The copy_sections remains to be separated.

One more question is that is copying objects to the same parent make sense? Do we have to support for example copying blk1.data_array[0] to blk1? Now is not supported.

jgrewe:
@hkchekc I see your point with the overload, also not sure, which version is better, but I can see several benefits in this version.
Regarding your point on blk1.data_arrays[0] to blk1: There would be an error on trying this because names must be unique within one group. But, on the other hand, it might be really useful if I want to create a new dataArray based on the info in the one passed via `copy_from`. In this case the overload method might come in handy.

Let's say I have a data_array within one block. Now I want another one with the same specifications but a different name:
```
blk1.create_data_array(name="new_array", copy_from=blk.data_arrays[0])
```
If I pass a name, type or other argument, it will replace the one from the template. Does that make sense?

hkchekc:
@jgrewe @achilleas-k 

I think what @jgrewe suggested make sense and I added such possibility.

Issue #368, opened 2019-01-24
Consistently update modification times


achilleas-k:
NIX Objects contain `created_at` and `updated_at` properties. The latter isn't always updated when changes are made to an object. We should update any setter methods to update the timestamp.

hkchekc:
There is a few difference with the NIX and NIXpy timestamp updates.

In NIXpy the force_updated is only called when:

- MultiTag - Positions
- Tag - Unit
- Feature - Data
- Creation of Property

In C++ NIX, forceUpdate is called for:
- the Four setters above
- SampledDiemnsions - Label
- Section - repository
- DataArray - Labels

Except for the setters, there are also case where the property is changed somewhere. For example, the dimensions in DataArray. In this case, whenever the dimension is appended, the DataArray should also be forced update.

jgrewe:
@hkchekc, good that you had a look into this! This is a long standing issue.

I am not entirely sure, how to handle this. On the one hand it should be updated whenever something is changed. On the other hand we have the case in relacs that we continuously write data or append positions stored in DataArrays and MTags. In these updates time matters and I don't really like the idea of always updating the modification time. That is, we may want to put this responsibility entirely into the hands of the user (with the exception that it is automatically set upon entity creation), or we provide an option/flag to specify if an update is wanted.

achilleas-k:
Interesting case.  I hadn't considered that.  I'd be curious to see how big the impact would be.

As an alternative, I wonder if we could somehow keep modification times in memory and flush them on close, but I think that would be too much work.  Your idea of somehow disabling mtimes for long sessions is definitely more straightforward.

jgrewe:
not sure about the impact myself, but the usual size of data chunks written are below 1024 samples that means that the update is executed about 20 times per second. If I am not mistaken, our modification time does not include milliseconds, does it?

achilleas-k:
Nope.  Seconds only.

hkchekc:
@jgrewe  I think for some properties which will be changed frequently, like positions . We can do the way you suggest by putting a flag on it.

And for some other properties, like the lables or units or expansion origin, which is not supposed to append or change too frequently, I guess we can force it to update whenever it changes?

hkchekc:
@jgrewe @achilleas-k 

I am not sure about what kind of decision should be given to the users regarding this update.

[...]

PR #385, opened 2019-04-11
Add auto-time-update flag to NIXpy file

hkchekc:
Refer to the issue #368 

I added a flag for timestamp updating when the file is opened. 2 functions (time_auto_update_on() and time_auto_update_off()) are added for altering the flag after the file is opened.

The current policy is to update the time whenever a property of an Entity or File is changed and when the time_auto_update flag is True. I do not put the flag in the force_update_at() function because it is more flexible this way. Note that the update of metatdata will not change the update_at() time anyway.

The force_update_at() in creation is not affected.

As of the performance issue, the time used for changing the expansion_origin for data_array 100,000 times is:

**EDITED:**

always update (without a flag): around 45s
flag = True: around 45s (possibly slower for sections or property)
flag = False: around 20s 
do not provide an auto_update possibility at all: around 20s (slightly less than flag = False)

### DIFF
--- a/nixio/file.py
+++ b/nixio/file.py
@@ -60,7 +60,6 @@ class FileMode(object):
     ReadWrite = 'a'
     Overwrite = 'w'
 
-
 def map_file_mode(mode):
     if mode == FileMode.ReadOnly:
         return h5py.h5f.ACC_RDONLY
@@ -86,7 +85,7 @@ def make_fcpl():
 class File(object):
 
     def __init__(self, path, mode=FileMode.ReadWrite,
-                 compression=Compression.Auto):
+                 compression=Compression.Auto, auto_update_time=False):
         """
         Open a NIX file, or create it if it does not exist.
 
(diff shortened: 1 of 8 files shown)


achilleas-k:
I don't think the auto-update setting should be stored in the file.  This should be a session option, much like the compression flag and the open mode.

I also don't see the reason for having two methods, one for turning it off and one for on.  Why not a property `nixfile.auto_update_timestamps = {True,False}`?

nixio/file.py
```diff
@@ -200,6 +202,22 @@ def format(self):
...
+        """
+        A user defined flag which decided if time should always be updated
+        when properties are changed.
+
+        :type: bool
+        """
+        return self._time_auto_update
+
+    @time_auto_update.setter
+    def time_auto_update(self, bool):
```
achilleas-k: Can you change the name of the parameter here?  `bool` is the buit-in name for the bool type.

hkchekc: changed

achilleas-k:
Failing test is due to mismatching File Format versions with NIX breaking a compatibility test.  Will be fixed by #386.

Issue #382, opened 2019-04-02
Add 'append' and 'extend' methods for Property.values


mpsonntag:
Currently an existing `Property.value` cannot be modified and has to be replaced, if a value is added.

`append` and `extend` methods for `Property.values` would make working with them more convenient.

achilleas-k:
Added in #392

PR #392, opened 2019-05-07
[Property values] Add append_values method for modifying values

hkchekc:
Refer to Issue #382.

### DIFF
--- a/nixio/property.py
+++ b/nixio/property.py
@@ -115,15 +115,6 @@ class Property(Entity):
 
         return newentity
 
-    def append_values(self, data):
-        """Method for appending values to existing values"""
-        arr = np.array(data).flatten('C')
-        ds = self._h5dataset.dataset
-        src_len = len(self.values)
-        dlen = len(arr)
-        ds.resize((src_len+dlen,))
-        ds.write_direct(arr, dest_sel=np.s_[src_len: src_len+dlen])
-
     @property
     def name(self):
         return self._h5dataset.get_attr("name")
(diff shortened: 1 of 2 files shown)


jgrewe:
I guess we need a test case for that :)

nixio/property.py
```diff
@@ -115,6 +115,15 @@ def _create_new(cls, nixparent, h5parent, name, dtype, oid=None):
 
         return newentity
 
+    def append_values(self, data):
+        """Method for appending values to existing values"""
+        arr = np.array(data).flatten('C')
+        ds = self._h5dataset.dataset
```
achilleas-k: H5DataSet exposes the `resize()` method through the `shape` property setter:
https://github.com/G-Node/nixpy/blob/ef2637cd198ef98b936e0c1211adbf76d8b57943/nixio/hdf5/h5dataset.py#L74

achilleas-k:
I was wondering if there was a more pythonic way to do this and I realised that, since `values` is a property with a getter and a setter, anything we do on the property directly should just work, assuming the operation works on the type returned by the getter.

So this doesn't work:
```python
prop.values.append(10)
```
because `values` is a tuple and tuples don't have an append method, but this does
```python
prop.values += (10,)
```
because you can concatenate tuples with `+`.

It's essentially equivalent to
```python
v = prop.values
v += (10,)
prop.values = v
```
and our `values` setter already takes care of resizing when necessary.
https://github.com/G-Node/nixpy/blob/ef2637cd198ef98b936e0c1211adbf76d8b57943/nixio/property.py#L273

I suppose it might still be useful for users to have an `append_values()` method.  Though I think it could be simplified to just a `+` operation on the values property.  Perhaps we should also call it `extend()` if it's going to support multi-value append.  Could do both.

nixio/property.py
```diff
@@ -115,6 +115,15 @@ def _create_new(cls, nixparent, h5parent, name, dtype, oid=None):
 
         return newentity
 
+    def append_values(self, data):
+        """Method for appending values to existing values"""
```
achilleas-k: The docstring shouldn't say what the method does, not describe what it is.

E.g., "Appends values to existing data."

nixio/property.py
```diff
@@ -285,6 +276,38 @@ def values(self, vals):
...
+                raise TypeError("Array contains inconsistent values. "
+                                "Only values of type '{}' can be "
+                                "appended".format(vtype))
+
+        arr = np.array(data, dtype=vtype).flatten('C')
+        ds = self._h5dataset
+        src_len = len(self.values)
+        dlen = len(arr)
+        ds.shape = (src_len+dlen,)
+        ds.dataset.write_direct(arr, dest_sel=np.s_[src_len: src_len+dlen])
```
achilleas-k: We should probably expose the `write_direct()` method from the `h5py.Dataset` class through our `H5DataSet` wrapper.

Doing a `write_direct()` this way is good, more efficient than reading in the existing values, appending the new ones, and writing them back.  I don't like that we're duplicating the type checking code from the values setter property though.  We should probably pull those out and make a reusable private method.

Issue #374, opened 2019-03-05
Pretty print for data


achilleas-k:
Like metadata, would be nice to have pretty print for the data tree.

JiriVanek:
I would appreciate if pprint will be available also for "file"

PR #393, opened 2019-05-07
Add pprint function for file

hkchekc:
Refer to Issue #374.

Also adjusted section.pprint() a little bit

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -363,6 +363,18 @@ class Block(Entity):
         return finders._find_sources(self, filtr, limit)
 
     def pprint(self, indent=2, max_length=120, extra=True, start_depth=0):
+        """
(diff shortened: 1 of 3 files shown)


nixio/block.py
```diff
@@ -362,8 +362,12 @@ def find_sources(self, filtr=lambda _: True, limit=None):
             limit = maxint
         return finders._find_sources(self, filtr, limit)
 
-    def pprint(self, indent=2, max_length=120, extra=True):
-        print(self)
+    def pprint(self, indent=2, max_length=120, extra=True, file=False):
+        if file:
```
achilleas-k: If the `file` argument only controls indent, and the indent can be set from the calling function, why can't not increment the indent from the `file.pprint()` caller?

PR #394, opened 2019-05-14
Some changes to DataFrame API

hkchekc:
- Previously, setting only part of the units to None will raise Error, the problem is solved
- Print_table method now support printing the part of DataFrame
- read_columns method now has an option to group return values in different way

### DIFF
--- a/nixio/data_frame.py
+++ b/nixio/data_frame.py
@@ -268,15 +268,15 @@ class DataFrame(Entity, DataSet):
             col_sl = np.s_[:]
         cl = np.array(self.column_names)[col_sl]
         row_form = "{:^10}" * (len(cl) + 1)
-        print(row_form.format(" ", *cl))
+        print(row_form.format("column:", *cl))
         if self.units is not None:
-            print(row_form.format("unit", *self.units[col_sl]))
+            print(row_form.format(" unit:", *self.units[col_sl]))
         if not isinstance(row_sl, Iterable):
             ridx = list(range(len(self)))[row_sl]
         else:
             ridx = row_sl
         for i, row in enumerate(self._read_data(sl=row_sl)[list(cl)]):
-            print(row_form.format("Data{}".format(ridx[i]), *row))
+            print(row_form.format("  [{}]:".format(ridx[i]), *row))
 
     def _find_idx_by_name(self, name):
         for i, n in enumerate(self.column_names):
--- a/nixio/data_frame.py
+++ b/nixio/data_frame.py
@@ -200,7 +200,6 @@ class DataFrame(Entity, DataSet):
         get_row = self._read_data(sl=(index,))
         return get_row
 
-    # TODO: allow writing multiple cells at the same time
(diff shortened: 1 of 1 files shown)


nixio/data_frame.py
```diff
@@ -137,10 +140,20 @@ def read_columns(self, index=None, name=None, sl=None):
...
+        if group_by_cols:
+            gcol = []
+            for n in name:
+                get_col = self._read_data(sl=slic)[n]
+                get_col = [i for i in get_col]
+                gcol.append(get_col)
+            return gcol
+        else:
+            get_col = self._read_data(sl=slic)[name]
+            return get_col
```
achilleas-k: Having a switch for col-major vs row-major ordering is nice, but the two options return different types, which isn't very friendly.

Row-major order (default) returns a numpy.array while col-major order returns a list.

nixio/data_frame.py
```diff
@@ -254,18 +254,27 @@ def read_cell(self, position=None, col_name=None, row_idx=None):
...
+            row_sl = np.s_[:]
+        if col_sl is None:
+            col_sl = np.s_[:]
+        cl = np.array(self.column_names)[col_sl]
+        row_form = "{:^10}" * (len(cl) + 1)
+        print(row_form.format(" ", *cl))
+        if self.units is not None:
+            print(row_form.format("unit", *self.units[col_sl]))
+        for i, row in enumerate(self._read_data(sl=row_sl)[list(cl)]):
             print(row_form.format("Data{}".format(i), *row))
```
achilleas-k: This is from before this PR, but I don't think the `Data` prefix to the output row numbers is necessary or even very meaningful.  It's also ambiguous: It's the output row index, which is useful, but it can easily be mistaken for the data row index.  Not sure what the best solution here would be.  I can imagine either being useful (or both).

achilleas-k:
Looks good.  Just a couple of minor comments.  Not sure how to address the second one so maybe we can discuss, but for the first one the return type should definitely be consistent.

nixio/data_frame.py
jgrewe: agreed

nixio/data_frame.py
```diff
@@ -270,8 +271,12 @@ def print_table(self, row_sl=None, col_sl=None):
         print(row_form.format(" ", *cl))
         if self.units is not None:
             print(row_form.format("unit", *self.units[col_sl]))
+        if not isinstance(row_sl, Iterable):
+            ridx = list(range(len(self)))[row_sl]
+        else:
+            ridx = row_sl
         for i, row in enumerate(self._read_data(sl=row_sl)[list(cl)]):
-            print(row_form.format("Data{}".format(i), *row))
+            print(row_form.format("Data{}".format(ridx[i]), *row))
```
achilleas-k: Agreed.  Original (in-data) Row index is probably the most useful.  I still don't like that they're labeled `Data0, Data1, Data2,... DataN`.  An indication in the heading explaining what the number is would be more appropriate.  As it stands, `DataN` doesn't really mean much.

PR #395, opened 2019-05-14
Change base class of LinkType and DimensionType to Enum

hkchekc:
Issue #186.

The second commit is about DimensionType. Not sure if this make sense.

### DIFF
--- a/nixio/validate.py
+++ b/nixio/validate.py
@@ -146,7 +146,7 @@ class Validate:
 
         if da.dimensions:
             for i, dim in enumerate(da.dimensions):
-                if dim.dimension_type == 'range':
+                if dim.dimension_type.value == 'range':
                     try:
                         if len(dim.ticks) != da.data_extent[i]:
                             # if data_extent is used instead of len()
@@ -159,7 +159,7 @@ class Validate:
                         raise IndexError("Dimension of Dataarray and "
                                          "Number of Dimension object Mismatch")
 
-                if dim.dimension_type == 'set':
+                if dim.dimension_type.value == 'set':
                     # same as above
                     try:
                         if len(dim.labels) != da.data_extent[i]:
@@ -221,7 +221,7 @@ class Validate:
                 unit_list = [un for un in unit_list if un]
                 dim_list = [dim for refer in tag.references
                             for dim in refer.dimensions
-                            if dim.dimension_type != 'set']
+                            if dim.dimension_type.value != 'set']
                 if len(unit_list) != len(dim_list):
                     tag_err_list.append(
                         "Some dimensions of references have no units"
@@ -388,7 +388,7 @@ class Validate:
         elif not all(r_dim.ticks[i] <= r_dim.ticks[i+1]
                      for i in range(len(r_dim.ticks)-1)):
             rdim_err_list.append("Ticks are not sorted!")
-        if r_dim.dimension_type != 'range':
+        if r_dim.dimension_type.value != 'range':
             rdim_err_list.append("Dimension type is not correct!")
 
         # sorting is already covered in the dimensions.py file
@@ -412,7 +412,7 @@ class Validate:
             da = self.errors['blocks'][blk_idx]['data_arrays'][da_idx]
             da['dimensions'][dim_idx]['errors'].append(self.check_dim(set_dim))
             self.error_count += 1
-        if set_dim.dimension_type != 'set':
+        if set_dim.dimension_type.value != 'set':
             da = self.errors['blocks'][blk_idx]['data_arrays'][da_idx]
             dim = da['dimensions'][dim_idx]
             dim['errors'].append("Dimension type is not correct!")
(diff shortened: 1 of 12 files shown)


achilleas-k:
This needs more work.  The Enum features aren't fully utilised and it makes working with them less convenient than what we had before.

For instance, if we change the string values of the LinkType to all lowercase, we no longer need the util functions for converting back and forth.  `util.link_type_from_string(str)` simply becomes `LinkType(str)` and `util.link_type_to_string(LinkType)` can then be replaced by `LinkType.value`.

Similarly, if the `Dimension.dimension_type` property returns an enum (`DimensionType(self._h5group.get_attr("dimension_type")`) instead of the string, then all the old comparisons can stay the same, instead of comparing with the value.

hkchekc:
@achilleas-k 

In general, this is a good idea and I did it that way. With the exception of validator because I don't want to import DimensionType there.

I also did a small change to the setter of link_type and dimension_type. I think allowing the use of strings and the strings with "wrong"-case will be better.

achilleas-k:
@gicmo @jgrewe I just had a chat with @hkchekc and I think it would be nicer if we enforce enums everywhere.  For instance, when creating a feature, this would make:
```python
tag.create_feature(da, "Untagged")
```
raise an error.  Instead, one should always use
```python
tag.create_feature(da, nix.LinkType.Untagged)
```

How do you feel about this?

jgrewe:
@achilleas-k, @hkchekc, yes, full agreement from my side. Would you prefer to immediately raise an error, when a string is passed, or would you be forgiving and only deprecate the string usage?

nixio/validate.py
```diff
@@ -146,7 +146,7 @@ def check_data_arrays(self, da, da_idx, blk_idx):
 
         if da.dimensions:
             for i, dim in enumerate(da.dimensions):
-                if dim.dimension_type == 'range':
+                if dim.dimension_type.value == 'range':
```
gicmo: Would it be more idiomatic to do `dim.dimension_type == DimensionType.Range`?

gicmo:
> Would you prefer to immediately raise an error, when a string is passed, or would you be forgiving and only deprecate the string usage

From a a quick glance, e.g. at the `Feature.link_type` setter, currently it would fail with something like `AttributeError: 'str' object has no attribute 'value'`. Which is not super descriptive. Maybe it would be good to wrap these places with `LinkType(link_type)` calls so we get better errors; supporting the old strings would be a nice by-product. I did a quick check and calls to `Enum(x)` seems to be idempotent anyway:
```python3
class test(Enum):
    foo = 'haha'
    bar = 'hehe'

x = test('haha')
y = test(x)

id(x) == id(y)
 → True
```

nixio/validate.py
achilleas-k: In general, one should be able to change the value of the enum with no (or little) consequence.  In our case, the consequence is the value stored on disk would change (breaking the NIX format), but all comparisons would work.

achilleas-k:
> ...
>  I did a quick check and calls to `Enum(x)` seems to be idempotent anyway:
> ... 
> `neat code snippet`

Oh, that's pretty cool.  Didn't know.

nixio/validate.py
gicmo: I guess we should add a comment to the enums as well, that the values of the enum are part of the NIX file format.

achilleas-k: Good idea!

achilleas-k:
All the comments have been addressed.  Merging!

hkchekc:
> > Would you prefer to immediately raise an error, when a string is passed, or would you be forgiving and only deprecate the string usage
> 
> From a a quick glance, e.g. at the `Feature.link_type` setter, currently it would fail with something like `AttributeError: 'str' object has no attribute 'value'`. Which is not super descriptive. Maybe it would be good to wrap these places with `LinkType(link_type)` calls so we get better errors; supporting the old strings would be a nice by-product. I did a quick check and calls to `Enum(x)` seems to be idempotent anyway:
> 
> ```python
> class test(Enum):
>     foo = 'haha'
>     bar = 'hehe'
> 
> x = test('haha')
> y = test(x)
> 
> id(x) == id(y)
>  → True
> ```

@gicmo while the idea is great and being merged, I just want to note that the by-product (supporting old strings) is not going to work (at least not completely). In the past, the string  'Tagged' with the first letter capitalized works. Now the valid string parameter input should be all lower-case. For example, 'tagged'

gicmo:
> @gicmo while the idea is great and being merged, I just want to note that the by-product (supporting old strings) is not going to work (at least not completely). In the past, the string 'Tagged' with the first letter capitalized works. Now the valid string parameter input should be all lower-case. For example, 'tagged'

Ah true, I guess for that a `.lower()` call would need to be added now. I have no strong opinion about supporting it (or not). @jgrewe what do you think?

jgrewe:
I would vote for supporting it, and thus to add the `.lower()` call

achilleas-k:
Issue created as a reminder.

Issue #364, opened 2019-01-24
Add ability to read old-style metadata to v1.5


achilleas-k:
Old style metadata (`Value` object) should be readable in the new library. The new file format should be upgraded to `1.2.0`. Library-file compatibility would then be as follows:
- NIX `1.4.x` will be unable to read `1.2.0` files
- NIX `1.5.x` will be able to read and write `1.2.0` files
- NIX `1.5.x` will be be able to read (not write) `1.1.1` files

matham:
I'm wondering what this change means for how I use nixpy. I have been using 1.4.x and all my files are in it. Most of the people in my lab are also on 1.4.x.

I wanted to try 1.5.x so I installed the most recent 1.5.x beta. However, it immediately choked trying to load the files created on 1.4.x because of this change. Specifically, in 1.4.x I read a property with `prop.values[0].value`, but with 1.5.x I need to do `prop.values[0]`.

My question is,
* if I create a file in 1.5.x, would I be able to read it in 1.4.x? I'm guessing not.
* if I create a file in 1.4.x, presumably once this issue is fixed I'll be able to read it in 1.5.x?
* For me it seems like creating a special function which reads the property depending on the nix version is really the only option to be able to support both 1.5.x and 1.4.x!?
* Is there a changelog somewhere that I could look for other breaking changes between versions? I tried looking for one, but couldn't find it.

jgrewe:
Hi @matham, you are right, that the 1.4.x libraries will not be able to read files created with 1.5.x libraries. The 1.5.x libraries will be able to read (but not write) files written with 1.4.x libs. We do not maintain a dedicated changelog (but probably should do so). Changes are always indicated in the release notes, though.
We are not done with the 1.5 version yet (still beta). We will make sure the release notes contain the needed information (in particular, the breaking ones).  There will be a converter script.

* With 1.5 we introduce ``DataFrames`` that allow storing spreadsheet like data. 
* The c++ bindings were dropped. nixpy is then a pure python package.
* We further simplified the metadata part of the nix data model to unify it with the odml data model.  This also simplifies handling the values, they are lists then.

The change you indicated should be mostly all that needs to be done when reading 1.4 files. Would you be fine with this? Do you have special demands? Please let us know and we will be happy to make the change as comfortable as possible.

matham:
Thanks for clarifying these points. The new `DataFrame` and being a pure python package sounds amazing.

For the reading part, I made this functions:

```py
[...]

PR #399, opened 2019-05-21
[prop] Add function to read old Values (incl. uncertainty, reference)

hkchekc:
Refer to Issue #364 
Originally, when reading values (getting props.values) from file with version < 1.1.1, it will return an old style compound type with (v.value, v.uncertainty, v.reference, v.filename, v.encoder, v.checksum). 

I break it down for value, uncertainty and reference. All now returns a tuple for values, single value for reference and uncertainty.

The problem is we cannot distinguish between an empty str or 0.0 and an unsetted uncertainty or reference.

### DIFF
--- a/nixio/property.py
+++ b/nixio/property.py
@@ -147,7 +147,7 @@ class Property(Entity):
     def uncertainty(self):
         dataset = self._h5dataset
         x, y, z = dataset._parent.file.attrs["version"]
-        if y < 1 or z < 1:
+        if x < 1 or (x == 1 and y < 1) or (x == 1 and y == 1 and z < 1):
             val = self._h5dataset.dataset[:]
             v = val[0]["uncertainty"]
             return v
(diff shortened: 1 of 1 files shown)


nixio/property.py
```diff
@@ -145,6 +145,12 @@ def unit(self, new):
 
     @property
     def uncertainty(self):
+        dataset = self._h5dataset
+        x, y, z = dataset._parent.file.attrs["version"]
+        if y < 1 or z < 1:
```
achilleas-k: Best to check if the entire ver tuple is `< (1, 1, 1)`.

nixio/property.py
```diff
@@ -215,9 +227,21 @@ def odml_type(self, new_type):
 
         self._h5dataset.set_attr("odml_type", str(new_type))
 
+    def _read_old_values(self):
+        val = self._h5dataset.dataset[:]
+        val_tu = tuple()
+        for v in val:
+            v = v["value"]
+            val_tu += (v,)
+        return val_tu
```
achilleas-k: @jgrewe @gicmo What's the current situation in NIX with reading old-style values and is there any plan to change it?  Do we ignore the extra information and just return a _new style_ tuple of raw values?

nixio/property.py
hkchekc: resolved

nixio/property.py
jgrewe: at the moment we just do that, returning a vector of Variant, i.e. the raw values, reference was swept under the rug, methinks...
need to remedy that

achilleas-k: We can go ahead and merge then and see how it develops in NIX.

PR #401, opened 2019-05-22
[docs] Reordering and minor rewrite of install page

achilleas-k:
- Moved installation instructions to the top since it's the most
important part and, for most users, the only piece of information they
care about.
- Dependencies moved to the end.  Users don't need this information
unless they're building from source (and even then, the pip method would
install them automatically).
- Changed dependency URLs to their PyPI page.  If users need to install
these manually, the PyPI page is more relevant.
- Added note to dev/beta version installation instructions that this
version should be used for testing new features.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -2,53 +2,51 @@
 Installation Guidelines
 =======================
 
+Installation instructions
+=========================
+
+The latest stable release of NIXPy is available on PyPi as `nixio`_.
+The simplest way to install NIXPy is to use pip::
+
+    pip install nixio
+
+Install developer version
+---------------------------
+The latest unstable (beta) version of NIXPy at any given time is also on PyPi and can be installed with pip::
+
+  pip install nixio --pre
+
+*Note that this version may contain untested code and should be used for testing new features in the format.*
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -2,53 +2,51 @@
...
-
-
 Installation instructions
 =========================
 
 The latest stable release of NIXPy is available on PyPi as `nixio`_.
-Therefore, the simplest way to install NIXPy is to use pip::
+The simplest way to install NIXPy is to use pip::
 
     pip install nixio
```
jgrewe: do we want to add the `pip3 install nixio` ?

achilleas-k: Good point.  Maybe I should also add which versions we support.  Something like
> NIXPy works with Python versions 2.7 and 3.6+ on Windows, macOS and Linux

And then we could have all pip commands duplicate
> ```
> pip3 install nixio
> ```
> or for Python 2.7
>```
> pip2 install nixio
> ```

Thoughts?

jgrewe: I think that would be helpful, indeed

PR #403, opened 2019-05-23
[GitHub actions] Add linting and test workflows

achilleas-k:
Adding workflow and Dockerfile for running pycodestyle for Python 3.7 on push.

I set up a GitHub Actions workflow on my fork for running `pycodestyle` for NIXPy.  I've opted the G-Node org into the GitHub Actions public beta and we should be getting the feature _soon_.  For now, it seems to be working [on my fork](https://github.com/achilleas-k/nixpy/runs/134737376) (note that the linting is actually failing; fix coming in a separate branch).

The workflows seem to be pretty straightforward to set up (though it took me a while to figure everything out).  The current workflow has two `action` steps.  The first (`build-pycodestyle37`) builds a Docker image, based on Python 3.7, which contains the code and the `pycodestyle` program.  The second action simply runs the container which triggers the entrypoint `pycodestyle /nixpy/nixio`.

The nice thing about these actions is that you can build and run the container locally before pushing and (if the workflows are defined correctly) it should run the same as on GH.  We probably shouldn't replace TravisCI just yet, but it's a nice extra bit of CI.


gicmo:
Just realized that the workflow file uses the old, deprecated format and probably be converted to the new YAML one.

achilleas-k:
Noticed that yesterday too.  I'll update with the new one.

gicmo:
> Noticed that yesterday too. I'll update with the new one.

Happy to merge it first and update it later, as you wish.

Issue #405, opened 2019-05-28
Fix references to _offset and _count in DataView


achilleas-k:
DataView contains leftover references to `_offset` and `_count` in its `write_data` function (see linked line).

https://github.com/G-Node/nixpy/blob/422e5c29720b42bb12bff3a2cdb580753e199272/nixio/data_view.py#L84

The DataView class no longer holds these two properties, but instead keeps a `slice`, which holds the equivalent to offset as `slice.start` and the equivalent to count as `slice.stop - slice.start` (normally this would require dividing by step, but in our case step is always 1).

We never test writing through a DataView so this wasn't caught in the rewrite for 1.5 when the bindings were removed.  We should add some tests.

hkchekc:
Same for https://github.com/G-Node/nixpy/blob/bb99824917d7d0bbeac9454516907097f9a773d5/nixio/multi_tag.py#L112

PR #406, opened 2019-05-28
[Type Checking] Add type annotations to all return from nixio method

hkchekc:
As the title suggested.

While adding notations, a few problems is seen, including Issue #405. Other problems includes:

- This pull request only works in python3.7 due to the from __future__ import annotations . If we drop this, we cannot check the class itself as an return type in the class.

- Mypy complains about signature of _create_new method of child class and Entity are not the same. I suggest to change _create_new to for example _create_new_data_array.  Same for any other override function which have incompatible return types with parent class.

- I suggest to force the data_type/ dtype properties of entites to return nix.DataType values

- Another Problem is how to check return types such as H5Group (in RangeDim), File (in Section) and other originally non-imported nix Classes.

- When checking for string type returns, is it better to use typing.AnyStr or simple str

 - Is it sufficient to check the return type is Container without checking what is inside. (Am not sure how to do it)


achilleas-k:
Changed the base to a new branch so that we don't break python2 in master.

nixio/dimensions.py
```diff
@@ -35,29 +36,29 @@ def _inst_item(self, item):
...
         return newdim
 
     @property
-    def dimension_type(self):
+    def dimension_type(self) -> DimensionType:
         return DimensionType(self._h5group.get_attr("dimension_type"))
 
     @dimension_type.setter
-    def _dimension_type(self, dimtype):
+    def _dimension_type(self, dimtype): -> None:
```
```suggestion
    def _dimension_type(self, dimtype) -> None:
```

PR #408, opened 2019-07-08
Support unicode in Property values

achilleas-k:
Fixed support for Unicode in Property values and added tests for Property and DataArray data.

### Property
When writing values, string types are run through ensure_text.  In Python 3 this converts to `str` and for Python 2 it converts to unicode.
When reading values, bytes are run through ensure_str.  In Python 3 this converts to str and for Python 2 it converts to basestring.

Added new test `test_unicode_values`.  The `ensure_str` conversion is replicated here to cover the case for Python2.

### DataArray
No write conversions are necessary since they're taken cafe of by the numpy array type.  The read conversion occurs in the same way as for the Property values, but in this case, it's returned as a numpy array with the vlen_str_dtype.

Added new test `test_array_unicode`.

### NIX Compatibility

Unicode property values and DataArray data are added to the NIX compatibility tests as well.

### DIFF
--- a/nixio/test/test_property.py
+++ b/nixio/test/test_property.py
@@ -191,7 +191,7 @@ class TestProperties(unittest.TestCase):
     def test_unicode_values(self):
         sec = self.section
         unistrings = {
-            "unlaut": ["ü"],
+            "umlaut": ["ü"],
             "gr": ["ω"],
             "deg": ["°"],
             "chinese-simplified": ["汉字"]
--- a/nixio/test/test_property.py
+++ b/nixio/test/test_property.py
@@ -186,3 +186,20 @@ class TestProperties(unittest.TestCase):
         number_extend = (1, 1.2)
         self.assertRaises(TypeError,
                           lambda: self.prop.extend_values(number_extend))
+
+    def test_unicode_values(self):
+        sec = self.section
+        unistrings = {
+            "unlaut": ["ü"],
+            "gr": ["ω"],
+            "deg": ["°"],
+            "chinese-simplified": ["汉字"]
+        }
+
+        # test writing
+        for name, value in unistrings.items():
+            sec.create_property(name, value)
+
+        # read them back
+        for name, value in unistrings.items():
+            assert tuple(value) == sec.props[name].values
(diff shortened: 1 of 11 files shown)


nixio/test/test_property.py
```diff
@@ -167,19 +168,40 @@ def test_empties(self):
...
+        self.assertRaises(TypeError,
+                          lambda: self.prop.extend_values(float_data))
+        number_extend = (1, 1.2)
+        self.assertRaises(TypeError,
+                          lambda: self.prop.extend_values(number_extend))
+
+    def test_unicode_values(self):
+        sec = self.section
+        unistrings = {
+            "unlaut": ["ü"],
```
jgrewe: this "Unlaut" is actually an "Umlaut"

jgrewe:
if I am not mistaken we still do not support unicode in units that are stored as hdf5 attributes. It would be a large overhead to store them as datasets, and a breaking change, so we leave it like that for now, correct?

nixio/test/test_property.py
achilleas-k: Woops

achilleas-k:
> if I am not mistaken we still do not support unicode in units that are stored as hdf5 attributes. It would be a large overhead to store them as datasets, and a breaking change, so we leave it like that for now, correct?

Lemme look into that right now.

achilleas-k:
> if I am not mistaken we still do not support unicode in units that are stored as hdf5 attributes. It would be a large overhead to store them as datasets, and a breaking change, so we leave it like that for now, correct?

Works for me
```python
import nixio as nix


nixfile = nix.File("testfile.nix", mode=nix.FileMode.Overwrite)
sec = nixfile.create_section("sec", "s")
temp = sec.create_property("temperature", 10.0)
temp.unit = "°C"

nixfile.close()

nixfile = nix.File("testfile.nix", mode=nix.FileMode.ReadOnly)
for p in nixfile.sections["sec"].props:
    print(p, p.values, p.unit)

nixfile.close()
```

```
❯ python nixunicodeunits.py
Property: {name = temperature} (10.0,) °C
```

achilleas-k:
Rebased on master again.  Waiting to check if there are any LGTM warnings or build failures.

PR #414, opened 2019-09-25
Fix DataView reading and writing

achilleas-k:
This PR fixes the read and write methods for DataView and adds tests for all sorts of weird cases and fancy slicing.

The rules for slicing and slice combinations were made to match Python slicing behaviour, adapted for the case of the DataView.  Most of the work is done in the `_transform_coordinates()` method.  An important change from the previous way the indices were calculated is that now we rely on `slice` objects and the `slice.indices()` method to calculate the start and stop values for fancy slicing cases, like negative indexing and slicing beyond the end of an array, both of which are now closer to the way indexing works in Python and NumPy.

New behaviour:
- Better handling of negative slices: `dataview[-N]` is legal only for `N < len(dataview)`.  The previous conversion simply calculated `len(dataview) - N` and then passed the result to the underlying h5py function, which raised the error when necessary.
- Slicing beyond the end of the array is now legal and returns the values up to and including the last one in the array.  For DataView objects, this stops at the internal `stop` value of the DataView.
- Negative step values are not allowed because HDF5 doesn't allow them for hyperslabs.  We catch this early to make stacktraces shorter.


achilleas-k:
Ok, now I'm done for real.

PR #421, opened 2019-10-12
Validator part 2

achilleas-k:
Did a bit of a cleanup and rewrite of the validator.

- The format of the results is a nested dictionary of the form
```python
results = {
    "errors": {
        nixobject: ["list of errors"]
    },
    "warnings": {
        nixobject: ["list of warnings"]
    }
}
```
- The validator doesn't throw any exceptions (like it did before), but it doesn't catch any read exceptions either.  I plan on adding this later so we can validate, for instance, a Block that's missing an ID.  Currently, the library throws an exception if an ID is missing.
- Error and warning strings are used from two static classes.  This puts the strings in a single location, so they're easier to change and test.
- Tests for every error and warning that doesn't break reading.

In a followup PR, I'll add a script so that the validator can be run from the command line.


jgrewe:
looks like we are caught in cycle here :)

nixio/validator.py
```diff
@@ -0,0 +1,498 @@
...
+            tag_errors, tag_warnings = check_tag(tag)
+            update_results(tag, tag_errors, tag_warnings)
+
+        # MultiTags
+        for mtag in block.multi_tags:
+            mtag_errors, mtag_warnings = check_multi_tag(mtag)
+            update_results(mtag, mtag_errors, mtag_warnings)
+
+        # Sources
+        def traverse_sources(sources):
```
jgrewe: is this on the right indentation level?

achilleas-k: I guess it could be outside the loop, but inside the outer method.

jgrewe: agreed, inside of ``check_file`` but outside of loop sounds good to me. Better than having method definition inside the loop, even though it will be usually only a few blocks in a file

achilleas-k:
I don't think the detected import cycles are real cycles.  Pycycle doesn't seem to think so.  I guess the issue detected is that `File` imports `validator` which imports `dimensions` and dimensions haven't been imported by `__init__.py` yet.

jgrewe:
> I don't think the detected import cycles are real cycles. Pycycle doesn't seem to think so. I guess the issue detected is that `File` imports `validator` which imports `dimensions` and dimensions haven't been imported by `__init__.py` yet.
> 
> EDIT: We could also drop the dimension type validation check since it's not really meaningful. The library initialises the Dimension object based on the `dimension_type` attribute when reading, so it's not like they can differ.

Could the dimension type be invalid? Not when the file was created via the library at least. I vote for skipping the check.

achilleas-k:
> Could the dimension type be invalid? Not when the file was created via the library at least. I vote for skipping the check.

It's the same as with other kinds of invalid states (like not having an ID).  The instantiation of the object will throw an exception, so there's way to have a Dimension object that's has an invalid dimtype.  I'll include it in the next version that will be able to catch exceptions and report them as errors.

PR #423, opened 2019-10-15
Validator script

achilleas-k:


gicmo:
> I don't have any objections against the name, btw.

A rose by any other name would smell as sweet. No objections either. (-:

Issue #420, opened 2019-10-01
ColumnDimension for linking DataArray dimension to a DataFrame column


achilleas-k:
A ColumnDimension on a DataArray would be similar to a SetDimension where the labels are taken from the column of a DataFrame.

Counterpart to G-Node/nix#748

PR #430, opened 2020-02-26
New DataArray dimension type: DataFrameDimension

hkchekc:
related to #420, #748

### DIFF
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -16,7 +16,8 @@ from .entity import Entity
 from .source_link_container import SourceLinkContainer
 from .datatype import DataType
 from .dimensions import (Dimension, SampledDimension, RangeDimension,
-                         SetDimension, DimensionType, DimensionContainer)
+                         SetDimension, DataFrameDimension,
+                         DimensionType, DimensionContainer)
 from . import util
 from .compression import Compression
 
@@ -140,6 +141,23 @@ class DataArray(Entity, DataSet):
         if self._parent._parent.time_auto_update:
             self.force_updated_at()
         return rdim
+
+    def append_data_frame_dimension(self, data_frame, column=None):
+        """
+        Append a new DataFrameDimension to 
+        :param data_frame:
+        :param column:
+        :return:
+        """
+        dimgroup = self._h5group.open_group("dimensions")
+        index = len(dimgroup) + 1
+        dfdim = DataFrameDimension._create_new(dimgroup, index,
+                                               data_frame)
+        if column:
+            dfdim.column = column
+        if self._parent._parent.time_auto_update:
+            self.force_updated_at()
+        return dfdim
 
     def append_alias_range_dimension(self):
         """
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -144,7 +144,8 @@ class DataArray(Entity, DataSet):
 
     def append_data_frame_dimension(self, data_frame, column=None):
         """
-        Append a new DataFrameDimension to 
+        Append a new DataFrameDimension to the list of existing dimension
+        descriptors.
         :param data_frame:
         :param column:
         :return:
@@ -152,9 +153,7 @@ class DataArray(Entity, DataSet):
         dimgroup = self._h5group.open_group("dimensions")
         index = len(dimgroup) + 1
         dfdim = DataFrameDimension._create_new(dimgroup, index,
-                                               data_frame)
-        if column:
-            dfdim.column = column
+                                               data_frame, column)
         if self._parent._parent.time_auto_update:
             self.force_updated_at()
         return dfdim
--- a/nixio/dimension_type.py
+++ b/nixio/dimension_type.py
@@ -15,3 +15,4 @@ class DimensionType(Enum):
     Sample = "sample"
     Range = "range"
     Set = "set"
+    DataFrame = "data_frame"
--- a/nixio/dimensions.py
+++ b/nixio/dimensions.py
@@ -321,7 +321,7 @@ class DataFrameDimension(Dimension):
         super(DataFrameDimension, self).__init__(h5group, index)
 
     @classmethod
-    def _create_new(cls, parent, index, data_frame):
+    def _create_new(cls, parent, index, data_frame, column):
         """
         Create new Dimension that point to a DataFrame
         :param parent: the linked DataArray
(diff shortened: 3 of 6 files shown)


nixio/dimensions.py
```diff
@@ -313,3 +313,46 @@ def labels(self):
...
+        newdim._dimension_type = DimensionType.DataFrame
+        return newdim
+
+    @property
+    def data_frame(self):
+        return self._h5group.get_attr("data_frame")
+
+    @data_frame.setter
+    def data_frame(self, df):
+        self._h5group.set_attr("data_frame", df)
```
achilleas-k: I think the DataFrame should be linked as a reference using `create_link()`, like we do with the AliasRangeDimension that references a DataArray (see below).
https://github.com/G-Node/nixpy/blob/4829ed7f9fe337c7d8595ad4d7d5edda349dc981/nixio/dimensions.py#L182

achilleas-k:
Could also add tests to this PR if we're happy with the decisions.  A few tests would have caught the lack of optional args for noted above.

achilleas-k:
Looks good.  Python 2 fails because of the ordered dict thing in the creation of DataFrame.  We do want to drop Py2 support, but I think for minor stuff we could probably still keep it going until v1.5.x.

I'd like to also see a test for the reading of the DataFrame through the Dimension, since the whole group instantiation thing might be a bit tricky.  In fact, test all the DataFrameDimension getters.  There's not that many of them.

nixio/test/test_dimensions.py
```diff
@@ -188,3 +188,35 @@ def test_append_dim_init(self):
...
+               (2, 'c', 20.05, 5.1, 100), (1, "d", 20.15, 5.3, 150),
+               (2, 'e', 20.23, 5.7, 200), (2, 'f', 20.07, 5.2, 300),
+               (1, "g", 20.12, 5.1, 39), (1, "h", 20.27, 5.1, 600),
+               (2, 'i', 20.15, 5.6, 400), (2, 'j', 20.08, 5.1, 200)]
+        unit =  [None, None, "s", "mV", None]
+        df = self.block.create_data_frame("ref frame", "test", col_dict=di, data=arr)
+        df.units = unit
+        dfdim1 = self.array.append_data_frame_dimension(df)
+        dfdim2 = self.array.append_data_frame_dimension(df, column=1)
+        self.assertRaises(ValueError,lambda: dfdim1.ticks())
```
achilleas-k:
```suggestion
        self.assertRaises(ValueError, lambda: dfdim1.ticks())
```

nixio/dimensions.py
```diff
@@ -313,3 +314,78 @@ def labels(self):
...
+    @property
+    def data_frame(self):
+        dfname = self._h5group.get_by_pos(0).name
+        grp = self._h5group.open_group(dfname)
+        df = DataFrame(grp.h5root.group["data_frames"], grp)
+        return df
+
+    @data_frame.setter
+    def data_frame(self, df):
+        self._h5group.create_link(df, df.id)
```
```suggestion
        self._h5group.create_link(df, "dataframe")
```

I think it might be simpler if the link to the DataFrame from the DataFrameDimension has a fixed name (e.g., `"dataframe"`), instead of linking by ID.  Just like with Features, Positions, and Extents, this would make it clearer what the purpose of the link is in the file and would make retrieving the referenced object simpler (see the `data_frame` property getter above).

@jgrewe Thoughts?

nixio/test/test_dimensions.py
```diff
@@ -188,3 +188,35 @@ def test_append_dim_init(self):
...
+                assert unit[idx] == dfdim2.unit(idx)
+                # labels
+                assert list(di)[idx] ==  dfdim1.label(idx)
+                assert list(di)[idx] ==  dfdim2.label(idx)
+        for ti, tu in enumerate(arr):
+            assert arr[ti][1] == dfdim2.ticks()[ti]
+        assert unit[1] ==  dfdim2.unit()
+        assert list(di)[1] ==  dfdim2.label()
+        assert dfdim1.label() == df.name
+
```
achilleas-k:
```suggestion
        assert self.array.dimensions[0] == df
```

achilleas-k:
Changes look good but the units indexing error is still uncaught (for when units are None).  Can you also rebase onto master?

nixio/dimensions.py
```diff
@@ -313,3 +315,81 @@ def labels(self):
...
+        :param column: the index of a column in the DataFrame that the Dimension will reference (optional)
+        :return: The new DataFrameDimension
+        """
+        newdim = super(DataFrameDimension, cls)._create_new(parent, index)
+        newdim.data_frame = data_frame
+        newdim.column = column
+        newdim._dimension_type = DimensionType.DataFrame
+        return newdim
+
+    def get_unit(self, index=None):
```
jgrewe: I think the public methods should have docstrings. Especially if using it with the default argument can lead to a ValueError.

nixio/dimensions.py
```diff
@@ -313,3 +315,81 @@ def labels(self):
...
+            else:
+                idx = self.column
+        else:
+            idx = index
+        unit = None
+        if self.data_frame.units is not None:
+            unit = self.data_frame.units[idx]
+        return unit
+
+    def get_ticks(self, index=None):
```
jgrewe: same here

nixio/dimensions.py
```diff
@@ -313,3 +315,81 @@ def labels(self):
...
+                                 "for this Dimension. Please supply one")
+            else:
+                idx = self.column
+        else:
+            idx = index
+        df = self.data_frame
+        ticks = df[df.column_names[idx]]
+        return ticks
+
+    def get_label(self, index=None):
```
jgrewe: and here :)

nixio/dimension_type.py
```diff
@@ -15,3 +15,4 @@ class DimensionType(Enum):
     Sample = "sample"
     Range = "range"
     Set = "set"
+    DataFrame = "data_frame"
```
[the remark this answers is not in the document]
jgrewe: true

nixio/data_array.py
```diff
@@ -141,6 +142,22 @@ def append_range_dimension(self, ticks, label=None, unit=None):
             self.force_updated_at()
         return rdim
 
+    def append_data_frame_dimension(self, data_frame, column=None):
+        """
```
jgrewe: Can you specify the arguments in the docstrings, In particular the meaning and caveats of providing a column or not. Also column is an index, right?

achilleas-k: Good point.  Perhaps `column_index`, although a bit long, would be a more self-explanatory argument name.

achilleas-k:
@hkchekc Would you like to add the docstrings to this PR?  Since Jan approved I can merge it and you can do them in another one if you want.  Let me know.

hkchekc:
@achilleas-k I think is good to merge this first

Issue #418, opened 2019-10-01
MultiTag: Convenience constructor that doesn't require DataArray


achilleas-k:
Counterpart to G-Node/nix#789

PR #436, opened 2020-03-06
Allow supplying np_arrays or list for multi_tags positions

hkchekc:
Also for extents.

Referring to #418, #789.

Also add some test for this constructor.

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -84,8 +84,6 @@ class Block(Entity):
 
         util.check_entity_name_and_type(name, type_)
         util.check_entity_input(positions)
-        if not isinstance(positions, DataArray):
-            raise TypeError("DataArray expected for 'positions'")
         multi_tags = self._h5group.open_group("multi_tags")
         if name in multi_tags:
             raise exceptions.DuplicateName("create_multi_tag")
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -56,8 +56,8 @@ class Block(Entity):
         return newentity
 
     # MultiTag
-    def create_multi_tag(self, name="", type_="", positions=0,
-                         copy_from=None, keep_copy_id=True):
+    def create_multi_tag(self, name="", type_="", positions=None,
+                         extents=None, copy_from=None, keep_copy_id=True):
         """
         Create/copy a new multi tag for this block.
 
@@ -83,11 +83,19 @@ class Block(Entity):
             return self.multi_tags[id]
 
         util.check_entity_name_and_type(name, type_)
-        util.check_entity_input(positions)
         multi_tags = self._h5group.open_group("multi_tags")
         if name in multi_tags:
             raise exceptions.DuplicateName("create_multi_tag")
-        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions)
+        if not isinstance(positions, DataArray):
+            da_name = "{}-positions".format(name)
+            positions = self.create_data_array(da_name, "{}-positions".format(type_),
+                                        data=positions)
+        if not isinstance(extents, DataArray) and extents is not None:
+            da_name = "{}-extents".format(name)
+            extents = self.create_data_array(da_name,
+                                        "{}-extents".format(type_)
+                                            , data=extents)
+        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions, extents)
         return mtag
 
     # Tag
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -86,16 +86,34 @@ class Block(Entity):
         multi_tags = self._h5group.open_group("multi_tags")
         if name in multi_tags:
             raise exceptions.DuplicateName("create_multi_tag")
-        if not isinstance(positions, DataArray):
-            da_name = "{}-positions".format(name)
-            positions = self.create_data_array(da_name, "{}-positions".format(type_),
-                                        data=positions)
-        if not isinstance(extents, DataArray) and extents is not None:
-            da_name = "{}-extents".format(name)
-            extents = self.create_data_array(da_name,
-                                        "{}-extents".format(type_)
-                                            , data=extents)
-        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions, extents)
+        try:
+            if not isinstance(positions, DataArray):
+                da_name = "{}-positions".format(name)
+                positions = self.create_data_array(da_name,
+                                                   "{}-positions".format(type_),
+                                                   data=positions)
+            if not isinstance(extents, DataArray) and extents is not None:
+                da_name = "{}-extents".format(name)
+                extents = self.create_data_array(da_name,
+                                                 "{}-extents".format(type_)
+                                                 , data=extents)
+            da_name = name
+            mtag = MultiTag._create_new(self, multi_tags,
+                                        name, type_, positions)
+        except (exceptions.DuplicateName, ValueError) as e:
+            msg = "Creation failed for {} due to {}.".format(da_name, e)
+            add_msg = None
+            if "{}-positions".format(name) in self.data_arrays:
+                del self.data_arrays["{}-positions".format(name)]
+                add_msg = " Created DataArray positions deleted."
+            if "{}-extents".format(name) in self.data_arrays:
+                del self.data_arrays["{}-extents".format(name)]
+                add_msg = " Created DataArray positions and extents deleted."
+            msg += add_msg
+            raise exceptions.CreationFail(msg)
+
+        if extents is not None:
+            mtag.extents = extents
         return mtag
 
     # Tag
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -103,7 +103,7 @@ class Block(Entity):
                 extcreated = True
             mtag = MultiTag._create_new(self, multi_tags,
                                         name, type_, positions)
-        except (exceptions.DuplicateName, ValueError) as e:
+        except Exception as e:
             msg = "MultiTag Creation Failed"
             if poscreated:
                 del self.data_arrays["{}-positions".format(name)]
@@ -111,7 +111,7 @@ class Block(Entity):
                 msg += " due to invalid positions"
             if extcreated:
                 del self.data_arrays["{}-extents".format(name)]
-            else:
+            elif poscreated and not extcreated:
                 msg += " due to invalid extents"
             print(msg)
             raise e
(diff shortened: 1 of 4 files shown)


btel:
# [Codecov](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=h1) Report
> Merging [#436](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=desc) into [master](https://codecov.io/gh/G-Node/nixpy/commit/b19e5ad4c875a8349f508d3f42c5acc3f7495501&el=desc) will **increase** coverage by `0.03%`.
> The diff coverage is `100.00%`.

[Impacted file tree graph](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #436      +/-   ##
==========================================
+ Coverage   90.41%   90.45%   +0.03%     
==========================================
  Files          55       55              
  Lines        7087     7114      +27     
==========================================
+ Hits         6408     6435      +27     
  Misses        679      679              
```


| [Impacted Files](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [nixio/hdf5/h5group.py](https://codecov.io/gh/G-Node/nixpy/pull/436/diff?src=pr&el=tree#diff-bml4aW8vaGRmNS9oNWdyb3VwLnB5) | `80.86% <0.00%> (-0.39%)` | :arrow_down: |
| [nixio/section.py](https://codecov.io/gh/G-Node/nixpy/pull/436/diff?src=pr&el=tree#diff-bml4aW8vc2VjdGlvbi5weQ==) | `80.97% <0.00%> (+0.07%)` | :arrow_up: |
| [nixio/property.py](https://codecov.io/gh/G-Node/nixpy/pull/436/diff?src=pr&el=tree#diff-bml4aW8vcHJvcGVydHkucHk=) | `71.74% <0.00%> (+0.25%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=footer). Last update [b19e5ad...41ceed4](https://codecov.io/gh/G-Node/nixpy/pull/436?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

nixio/multi_tag.py
```diff
@@ -52,7 +52,14 @@ def positions(self, da):
             raise TypeError("MultiTag.positions cannot be None.")
         if "positions" in self._h5group:
             del self._h5group["positions"]
-        self._h5group.create_link(da, "positions")
+        pos = da
+        if not isinstance(da, DataArray):
+            blk = self._parent
+            name = "f{self.name}-positions"
```
achilleas-k: Bad f-string.  It should be `f"{self.name}...`. 

More importantly, f-strings aren't supported in 2.7, so let's keep it Python2 friendly until after v1.5.

nixio/multi_tag.py
```diff
@@ -16,7 +16,7 @@
 from .data_view import DataView
 from .link_type import LinkType
 from .exceptions import (OutOfBounds, IncompatibleDimensions,
-                         UninitializedEntity)
+                         UninitializedEntity, DuplicateName)
```
achilleas-k: Added DuplicateName import but it's not used anywhere?

nixio/multi_tag.py
```diff
@@ -52,7 +52,14 @@ def positions(self, da):
...
         if "positions" in self._h5group:
             del self._h5group["positions"]
-        self._h5group.create_link(da, "positions")
+        pos = da
+        if not isinstance(da, DataArray):
+            blk = self._parent
+            name = "f{self.name}-positions"
+            if name in blk.data_arrays:
+                del blk.data_arrays[name]
+            pos = blk.create_data_array(name, "multi_tag_positions", data=da)
```
achilleas-k: Might be nicer if the type follows the same format as the naming: same type as the parent but with `"-positions"` or `".positions"` appended.

nixio/test/test_multi_tag.py
```diff
@@ -102,6 +102,19 @@ def tearDown(self):
         self.file.close()
         self.tmpdir.cleanup()
 
+    def test_multi_tag_new_constructor(self):
```
achilleas-k: Test should also check if DataArrays were created with the expected names.

achilleas-k: More importantly, check that there's two DataArrays with the expected names on the appropriate Block and add a check that those data arrays have the same ID as the position and extent references.

nixio/multi_tag.py
```diff
@@ -52,7 +52,15 @@ def positions(self, da):
...
             del self._h5group["positions"]
-        self._h5group.create_link(da, "positions")
+        pos = da
+        if not isinstance(da, DataArray):
+            blk = self._parent
+            name = "{}-positions".format(self.name)
+            if name in blk.data_arrays:
+                del blk.data_arrays[name]
+            pos = blk.create_data_array(name, "{}-positions".format(self.type),
+                                        data=da)
```
achilleas-k: What if `da` is an integer or string or some other invalid value?  I think it would be confusing for the user if they were trying to create a MultiTag and they got an error about DataArray creation with invalid data.  We should catch this issue as early as possible, preferably in the `create_multi_tag()` method in the `Block` class.

Also, I think this might not be the best location for the convenience functionality to be in general.  Having it here makes it possible to create DataArrays by assigning values to `mtag.positions`.  While this might sound useful, it also makes it more likely that a user tries to do something like:
```python
mtag.positions = [10, 11]
# ... later
mtag.positions = [10, 11, 12]
# or
mtag.positions += [13]
```
which will lead to all sorts of errors: Duplicate names for the DataArray creation, attempting to append a list to a DataArray, etc.

I think the convenience should only be in the `block.create_multi_tag()` method and nowhere else, and it should be documented clearly in the arguments in the docstring.  The type check for the `positions` argument in `create_multi_tag` should check for `DataArray` or any kind of `sequence`.

nixio/multi_tag.py
```diff
@@ -74,7 +82,16 @@ def extents(self, da):
...
+            ext = da
+            if not isinstance(da, DataArray):
+                blk = self._parent
+                name = "{}-extents".format(self.name)
+                if name in blk.data_arrays:
+                    del blk.data_arrays[name]
+                ext = blk.create_data_array(name,
+                                            "{}-extents".format(self.type)
+                                            , data=da)
+            self._h5group.create_link(ext, "extents")
```
achilleas-k: Given my comments above about not having the convenience functionality in the `positions` setter, I think the same should apply to `extents`.
I'm thinking we could have an `extents` argument in the `create_multi_tag()` method that works the same way as `positions`.

In both cases (for positions and extents), I think it only really makes sense to allow the user to automatically create DataArrays for MultiTag `positons` and `extents` **once** per MultiTag, on creation.  Otherwise, we're creating an incentive for users to create, recreate, and change the positions and extents through these creation functions and that would lead to losing track of DataArray creation, causing naming conflicts and other issues.

nixio/block.py
```diff
@@ -83,11 +83,19 @@ def create_multi_tag(self, name="", type_="", positions=0,
...
         util.check_entity_name_and_type(name, type_)
-        util.check_entity_input(positions)
         multi_tags = self._h5group.open_group("multi_tags")
         if name in multi_tags:
             raise exceptions.DuplicateName("create_multi_tag")
-        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions)
+        if not isinstance(positions, DataArray):
+            da_name = "{}-positions".format(name)
+            positions = self.create_data_array(da_name, "{}-positions".format(type_),
+                                        data=positions)
```
achilleas-k:
```suggestion
            positions = self.create_data_array(da_name, "{}-positions".format(type_),
                                               data=positions)
```

nixio/block.py
```diff
@@ -83,11 +83,19 @@ def create_multi_tag(self, name="", type_="", positions=0,
...
 
         util.check_entity_name_and_type(name, type_)
-        util.check_entity_input(positions)
         multi_tags = self._h5group.open_group("multi_tags")
         if name in multi_tags:
             raise exceptions.DuplicateName("create_multi_tag")
-        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions)
+        if not isinstance(positions, DataArray):
+            da_name = "{}-positions".format(name)
+            positions = self.create_data_array(da_name, "{}-positions".format(type_),
```
achilleas-k: Might be good to catch any DA creation exceptions here, specifically any exception that will be thrown if the `positions` data is invalid.  We could then re-raise it from here with a more appropriate message that mentions the MultiTag positions array explicitly.

nixio/block.py
```diff
@@ -83,11 +83,19 @@ def create_multi_tag(self, name="", type_="", positions=0,
...
+                                        data=positions)
+        if not isinstance(extents, DataArray) and extents is not None:
+            da_name = "{}-extents".format(name)
+            extents = self.create_data_array(da_name,
+                                        "{}-extents".format(type_)
+                                            , data=extents)
```
achilleas-k:
```suggestion
            extents = self.create_data_array(da_name,
                                             "{}-extents".format(type_),
                                             data=extents)
```

See also above comment about the DataArray creation exceptions.

nixio/block.py
```diff
@@ -83,11 +83,19 @@ def create_multi_tag(self, name="", type_="", positions=0,
...
+        mtag = MultiTag._create_new(self, multi_tags, name, type_, positions, extents)
```
achilleas-k: Don't change the `MultiTag._create_new()` method.  Just add extents to the mtag after it's created.

achilleas-k:
Looks good.  I'd like to see some more extensive tests.  Like having invalid positions (not just `None`, but like a single value), valid positions but invalid extents (and check if the positions DA was deleted when the exception is caught) etc.

Let me know if you want to include them here.  Otherwise rebase and I'll merge.

hkchekc:
Ok, then don't merge now. I will include them here

Issue #422, opened 2019-10-14
Create dimension iterator that returns 1-based index along with dimension


achilleas-k:
```
jgrewe 11:53
maybe a strange thought, but would be the 1.5 breaking event the right time, to change dimension index to start at 0? (edited) 
I still feel that there is a semantic bug in thinking of a dimension descriptor for the zeroth  dimension….
achilleas 12:20
Yeah, that's always a weird one
On the one hand is what you mentioned, on the other it's weird to be getting items from an indexable sequence using 1-based indexing
jgrewe 12:35
true, so you would be in favor of doing it zero_based?
and can we change it without feeling too bad?
achilleas 12:49
I don't know what I'm in favour of
gicmo 09:47
why did we do it like that back then?
jgrewe 09:48
because bad gut feelings about a zeroth dimension... (edited) 
gicmo 09:49
ah yes
heh, I personally don't care either way, ... so whatever you guys think
you that you are actually using it (-:
jgrewe 09:51
actually, I would enjoy 0-based indexing, in nixpy it is like that because ``da.dimensions`` is a list and in nixview I always have to do confusing conversions...
achilleas 14:16
I understand why 1-based makes some kind of sense for dimensions, but the issue is that there's no indications that da.dimensions[idx] is any different from any other collection that supports indexing (in the library or otherwise).
So making it 1-based creates this exception or difference in functionality that's not apparent and can only be understood in the context of the semantics of the property name.
I'd actually be in favour of having a method for getting a dimension in a 1-based indexy kind of way, but I can't think of a good name for it.
a sort of Java-like da.get_dimension(1)
The existence of the method might be enough to indicate that there's something different about this kind of access.
Though I could be fooling myself there
gicmo 14:36
yeah, although in C++ and python there are methods for [x] access, called get(x) where x is still 0 based
so a programmer might still be confused
achilleas 14:36
da.get_dimension_by_1_based_index(1)
Is this what they call self-describing method names?
gicmo 14:38
I guess that would work
:stuck_out_tongue_winking_eye:
achilleas 14:38
On a more serious note, we could have an iterator method that returns (dimindex, dimension) (similar to dict.items()) and dimindex would start at 1?
gicmo 14:39
that would indeed work well
achilleas 14:39
for idx, dim in da.iter_dimensions():
   ...
gicmo 14:39
and in C++ we would need some iterator based shit
could work
achilleas 14:39
I'm pasting this conversation in an issue
```

PR #438, opened 2020-03-09
Add method that returns 1-based index dimension iterators

hkchekc:
Refer to #422

### DIFF
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -217,6 +217,13 @@ class DataArray(Entity, DataSet):
         else:
             raise TypeError("Invalid Dimension object in file.")
 
+    def iter_dimensions(self):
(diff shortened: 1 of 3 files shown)


nixio/dimensions.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+
```
```suggestion
```

nixio/data_array.py
```diff
@@ -212,6 +213,11 @@ def _get_dimension_by_pos(self, index):
         else:
             raise TypeError("Invalid Dimension object in file.")
 
+    def iter_dimensions(self):
```
achilleas-k: Since this isn't a very standard method or behaviour, we should document it.

PR #446, opened 2020-03-17
Codestyle fixes

achilleas-k:
Followup to #444.
Codestyle fixes and unnecessary lambda removal.

A few snuck in while the previous PR was open.

@hkchekc Have a look at this PR and note the changes in the `assertRaises()` calls.  I went through and changed all the old ones I had written too so they don't use `lambda` since we can pass arguments to the callable.

There are a few `assertRaises()` calls that use lambda because they're testing indexing failures, so the arguments don't work exactly there.  We could also use `with self.assertRaises(<ErrorType>)`, but for now this should be good enough.

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -91,15 +91,15 @@ class Block(Entity):
         try:
             if not isinstance(positions, DataArray):
                 da_name = "{}-positions".format(name)
-                positions = self.create_data_array(da_name,
-                                                   "{}-positions".format(type_),
-                                                   data=positions)
+                positions = self.create_data_array(
+                    da_name, "{}-positions".format(type_), data=positions
+                )
                 poscreated = True
             if not isinstance(extents, DataArray) and extents is not None:
                 da_name = "{}-extents".format(name)
                 extents = self.create_data_array(da_name,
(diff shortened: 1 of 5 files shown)


nixio/block.py
```diff
@@ -91,15 +91,15 @@ def create_multi_tag(self, name="", type_="", positions=None,
         try:
             if not isinstance(positions, DataArray):
                 da_name = "{}-positions".format(name)
-                positions = self.create_data_array(da_name,
-                                                   "{}-positions".format(type_),
-                                                   data=positions)
+                positions = self.create_data_array(
+                    da_name, "{}-positions".format(type_), data=positions
+                )
```
jgrewe: you change the indentation here, but keep it the old style a few lines later :)
I am actually fine with both ways :man_shrugging:

achilleas-k: It's a line length issue, not so much indentation style.

I mean
```python
Foo(one,
    two,
    three)
```

and

```python
Foo(
    one,
    two,
    three
)
```
are both "valid" (conventionally).  I prefer the former, but sometimes the line length makes the latter necessary.

PR #451, opened 2020-03-25
File ID, header creation modifications, and File Format 1.2.0

achilleas-k:
Added ID in the file header.  Attribute key `id`.

Removed header property setters.  See corresponding commit for full explanation.

Updated file format version to 1.2.0.

PR #452, opened 2020-03-26
Upgrade script

achilleas-k:
This PR is mainly for the upgrade script but also includes some fixes and cleanup.

The script first checks the file header version to see if it's older than the current. If it is, it continues to collect tasks that need to be performed to complete the upgrade.

Tasks are created using functions that check if a change is required and if so, they return a closure bound to the specific filename. The returned function is then run after the user confirms they want to apply the changes. If --force is specified, the changes are applied without prompting.

All checks and changes are made in standalone functions that open the file using h5py and close it immediately after. This makes each change atomic (or atomic-ish) and reduces the chances that a file will be left in an invalid state. There's still a chance the user will stop the
procedure after a change has been applied and before the file version in the header is updated. This will put the file in a sort of invalid state but rerunning the upgrade script should finalise the changes.

Regardless of the possibilities of corruption or invalid states, the user is warned to make sure their files and data are backed up.

If you want to test the script, you can use the files in this GIN repository: https://gin.g-node.org/achilleas/nix-files.  This includes two files from the G-Node/nix-examples repository (which come from Jan) and two more I made, one using nixpy v1.4 and the other using v1.5 (current master).

### DIFF
--- a/nixio/cmd/upgrade.py
+++ b/nixio/cmd/upgrade.py
@@ -1,0 +1,19 @@
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Upgrade NIX files to newest version"
+    )
+    parser.add_argument("-f", "--force", action="store_true",
+                        help="overwrite existing files without prompting")
+    parser.add_argument("file", type=str, nargs="+",
+                        help="path to file to upgrade (at least one)")
+    args = parser.parse_args()
+    files = args.file
+
+    print(f"Upgrading {len(files)} files...")
+
+
+if __name__ == "__main__":
+    main()
--- a/nixio/cmd/upgrade.py
+++ b/nixio/cmd/upgrade.py
@@ -18,20 +18,27 @@ def has_valid_file_id(fname):
 
 def add_file_id(fname):
     """
-    Returns a closure that binds the filename. When the return value is
-    called, it adds a UUID to the file header.
+    Returns a closure that binds the filename if a file ID is required. When
+    the return value is called, it adds a UUID to the file header.
     """
+    if has_valid_file_id(fname):
+        return None
+
     def add_id():
         "Add a UUID to the file header"
         with h5py.File(fname, mode="a") as hfile:
+            if has_valid_file_id(fname):
+                return
+            print("Adding file id")
             hfile.attrs["id"] = nix.util.create_id()
     return add_id
 
 
 def update_property_values(fname):
     """
-    Returns a closure that binds the filename. When the return value is
-    called, it rewrites all the metadata Property objects to the new format.
+    Returns a closure that binds the filename if at least one Property update
+    is required. When the return value is called, it rewrites all the metadata
+    Property objects to the new format.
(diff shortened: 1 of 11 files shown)


achilleas-k:
Forgot to mention: This PR includes #451.

nixio/cmd/upgrade.py
```diff
@@ -0,0 +1,197 @@
...
+        "Add a UUID to the file header"
+        with h5py.File(fname, mode="a") as hfile:
+            hfile.attrs["id"] = nix.util.create_id()
+    return add_id
+
+
+def update_property_values(fname):
+    """
+    Returns a closure that binds the filename. When the return value is
+    called, it modifies rewrites all the metadata Property objects to the new
```
jgrewe: missing "and" like "modifies and rewrites"?

achilleas-k: OH NO!!

achilleas-k:
I had a thought about a potential edge case that might cause issues.  If you supply the same file twice, the task collection procedure will put the same file on the list twice.  This wont be a problem with the file ID and version update, but the property reformatting might cause issues.  Most likely the script will crash when attempting to get the "uncertainty" component of the property's value, so it shouldn't cause any corruption.

This will happen regardless of whether the file is specified twice on the command line, or if it's through a symlink or hardlink.

achilleas-k:
Tested now and as I suspected it crashes when trying to get the first field of the compound data type in the property.  I can imagine this might be annoying if someone is running an upgrade of all their files and they have a symlink to one somewhere.  I guess the chances are low in general, but maybe we should guard against it.

This isn't an issue if the same file is specified twice on the command line, since the tasks are indexed by the filename and that will get deduplicated.

achilleas-k:
With the latest changes, the script is a bit more careful about what it adds to the tasks and what it performs.  It will only add a task if it's needed (previously, the Property update task was added even if the number of Property objects that needed updating was 0) and before performing any change, it rechecks if it's needed.

nixio/cmd/upgrade.py
jgrewe: welcome

nixio/cmd/upgrade.py
```diff
@@ -0,0 +1,219 @@
...
+                if any(checksum):
+                    create_property(hfile, propname + ".checksum",
+                                    dtype=nix.util.vlen_str_dtype,
+                                    data=checksum)
+
+    update_props.__doc__ = "Update {} properties".format(len(props))
+    return update_props
+
+
+def create_property(hfile, name, dtype, data):
```
jgrewe: I am afraid the unit got lost during upgrading

achilleas-k: Oh dear.  Good catch!

achilleas-k:
I'll add an extra step at the end to offer to run `h5repack` on the upgraded files to reclaim the space from the deleted objects.  This will create a copy of each file, so the user should be warned of space usage (and potential process time).

PR #461, opened 2020-05-04
new command line tool for exploring nix files

jgrewe:
I find myself quite often in the situation that I am looking for a tiny bit of information in our recordings (such as the species, cell type, or recording quality) and then, instead of looking into the nix files, I start opening the old ascii files we still dump during recording.

This command line tool  allows for exploring the file and look for information. So far it is just a rough sketch but I would love to know what you think, your ideas and critique.

I was wondering whether this tool should be part of nixpy or rather nixworks. I came to the conclusion that it should be part of the main package...


nixio/cmd/nixinfo.py
```diff
@@ -31,41 +41,72 @@ def disp_metadata(filename, arguments):
...
-        description="Search for information within NIX file(s)"
-    )
-    parser.add_argument("file", type=str, nargs="+",
-                        help="Path to file (at least one)")
-    parser.add_argument("suffix", type=str, default="nix", nargs="?", 
-                        help="The suffix used for nix data files (default: nix).")
-    parser.add_argument("-m", "--metadata", type=str, default="",
-                        help="display file metadata, accepts a pattern")
-    args = parser.parse_args()
+    parser = argparse.ArgumentParser(prog="nixo-info",
```
achilleas-k:
```suggestion
    parser = argparse.ArgumentParser(prog="nixio-info",
```

nixio/cmd/nixinfo.py
```diff
@@ -63,18 +63,22 @@ def data_worker(arguments):
...
 def file_worker(arguments):
-
+    files = assemble_files(arguments)
+    for nf in files:
+        disp_file_info(nf, arguments)
     pass
 
 
-def main():
+def __create_parser():
```
achilleas-k:
```suggestion
def create_parser():
```

Why `__`?

jgrewe:
this should be ready for review for now. Some things are more sketches than full solutions. Some linting fails on the tool for too long line issues, but I am not sure if I would really like to break them down...

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,623 @@
...
+        print(e)
+        print("\n")
+    nix_file.close()
+
+
+def data_worker(arguments):
+    files = assemble_files(arguments)
+    func = disp_data
+    for nf in files:
+        func(nf, arguments)
```
achilleas-k, line 511:
```suggestion
    for nf in files:
        disp_data(nf, arguments)
```

This function name assignment seems unnecessary.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,623 @@
...
+    for filename in filenames:
+        if os.path.isfile(filename):
+            all_files.append(filename)
+        elif os.path.isdir(filename):
+            filename = os.sep.join((filename, "*." + arguments.suffix))
+            all_files.extend(sorted(glob.glob(filename)))
+        else:
+            candidates = sorted(glob.glob(filename))
+            if len(candidates) == 0:
+                print("Error: invalid file or directory! No matches found. '{}'").format(filename)
```
achilleas-k, line 95:
```suggestion
                print("Error: invalid file or directory! No matches found. '{}'".format(filename))
```

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,623 @@
...
+
+
+def mdata_worker(arguments):
+    files = assemble_files(arguments)
+    for nf in files:
+        disp_metadata(nf, arguments)
+
+
+def find_data_entity(nix_file, arguments):
+    name_or_type = arguments.pattern if arguments.case_sensitive else arguments.pattern.lower()
```
achilleas-k, line 332: Causes error if no pattern is supplied on the command line.  Should patterns be required or should this pretty-print the data tree when no patterns are supplied.  I'd go for printing the whole data tree for consistency with the `mdata` subcommand.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,623 @@
...
+        dump_threed(data, array.dimensions, array.label, array.unit)
+    else:
+        print("Sorry, cannot dump data with more than 3 dimensions!")
+
+
+def data_dump(filename, arguments):
+    nix_file = open_nix_file(filename)
+    entities = find_data_entity(nix_file, arguments)
+    for e in entities:
+        if isinstance(e, nix.pycore.data_array.DataArray):
```
achilleas-k, line 470:
```suggestion
        if isinstance(e, nix.DataArray):
```

`pycore` is gone.  Classes are exposed at the top level.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,623 @@
...
+            dump_data_array(e, filename)
+
+    nix_file.close()
+
+
+def data_plotter(filename, arguments):
+    nix_file = open_nix_file(filename)
+    entities = find_data_entity(nix_file, arguments)
+    for e in entities:
+        if isinstance(e, nix.pycore.data_array.DataArray):
```
achilleas-k, line 480:
```suggestion
        if isinstance(e, nix.DataArray):
```

nixio/cmd/nixexplore.py
jgrewe: actually, pattern expects one argument, if no argument is given, there will be an error raised by argparse. If you pass an empty string, it will print out a summary of all stored entities.

achilleas-k: I mean that the `-p` flag is optional but this line assumes it's not and tries to read it.  So if you run something like `nixio-explore dump <filename>` you get
```
Traceback (most recent call last):
File "/home/achilleas/.local/bin/nixio-explore", line 33, in <module>
sys.exit(load_entry_point('nixio', 'console_scripts', 'nixio-explore')())
File "/home/achilleas/projects/nixpy/nixio/cmd/nixexplore.py", line 622, in main
args.func(args)
File "/home/achilleas/projects/nixpy/nixio/cmd/nixexplore.py", line 523, in dump_worker
func(nf, arguments)
File "/home/achilleas/projects/nixpy/nixio/cmd/nixexplore.py", line 468, in data_dump
entities = find_data_entity(nix_file, arguments)
File "/home/achilleas/projects/nixpy/nixio/cmd/nixexplore.py", line 332, in find_data_entity
name_or_type = arguments.pattern if arguments.case_sensitive else arguments.pattern.lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```
because `-p` wasn't supplied and `arguments.pattern` is None.  We should either check for None or set the `-p` flag to have a default value of empty string.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,622 @@
...
+    add_default_file_args(file_parser)
+    file_parser.set_defaults(func=file_worker)
+
+
+def create_parser():
+    parser = argparse.ArgumentParser(prog="nixio-explore",
+                                     description=general_help)
+    subparsers = parser.add_subparsers(title="commands",
+                                       help="Sub commands for working on data and metadata",
+                                       description=tool_description)
```
achilleas-k, line 604: There's a small issue here with running `nixio-explore` without a subcommand.  Since it's not marked as required, the `func` attribute isn't set if no subcommand is specified and the `args.func(args)` in main throws an exception.

Adding `required=True` is one solution, but that was added in Python 3.7, so if we want to support 3.6 it's not an option.  The other solution is to add `dest="func"` to `add_subparsers()`, which will create the `func` attribute every time and then in main we can do something like
```python
    parser = create_parser()
    args = parser.parse_args()
    if not args.func:
        parser.print_help()
        sys.exit(1)
    args.func(args)
```

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,622 @@
...
+
+
+def get_dim_label_and_unit(dimension):
+    dim_label = getattr(dimension, "label") if hasattr(dimension, "label") else ""
+    dim_unit = getattr(dimension, "unit") if hasattr(dimension, "unit") else ""
+    return dim_label, dim_unit
+
+
+def dump_oned(data, dimension, label, unit, format="%.6f", end="\n\n"):
+    assert(len(data.shape) == 2)
```
achilleas-k, line 374: Shouldn't this be `== 1` for oned?

More generally though, Assertions in scripts are a little user hostile.  They'll produce an AssertionError and traceback which isn't very helpful.  A friendlier error message should be printed if this happens.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,622 @@
...
+    ticks = get_ticks(dimension, data.shape[0])
+    dim_label, dim_unit = get_dim_label_and_unit(dimension)
+    max_tick_len = max([len(format % ticks[-1]), len(dim_label)])
+
+    numeric_dim_ticks = isinstance(ticks[0], (int, float))
+    numeric_data = isinstance(data[0, 0], (int, float))
```
achilleas-k, line 380: 2-dimensional indexing of 1D data.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,622 @@
...
+    parent_parser.add_argument("-c", "--case_sensitive", action="store_true", help="matching of"
+                               + " entitiy names and types is case sensitive, by default the case is ignored")
+    parent_parser.add_argument("-fm", "--full_match", action="store_true", help="names and types must"
+                               + " be full matches, bey default a partial match is sufficient")
+
+
+def create_metadata_parser(parent_parser):
+    meta_parser = parent_parser.add_parser("metadata", help="Filter and display metadata", aliases=["mdata"],
+                                           description="Search for metadata items or display metadata (sub)trees.")
+    meta_parser.add_argument("-p", "--pattern", type=str, default=[], nargs="+", help=mdata_pattern_help)
```
achilleas-k, line 550: The `nargs="+"` here creates an issue where running something like:
```
nixio-explore metadata -p "pattern" <filename>
```
causes `<filename>` to be interpreted as a second pattern and then the command fails with
```
nixio-explore metadata: error: the following arguments are required: file
```

The user can get around this by instead calling
```
nixio-explore metadata -p "pattern" -- <filename>
```
but this isn't immediately obvious and we should find a better way to handle multiple patterns.

nixio/cmd/nixexplore.py
jgrewe: ah I see, was testing with the data subcommand

jgrewe: should be solved, pattern got defaults now

nixio/cmd/nixexplore.py
jgrewe: thanks, done

nixio/cmd/nixexplore.py
jgrewe: ups, fixed

nixio/cmd/nixexplore.py
jgrewe: using the "append" method now so multiple patterns are passed like `nixio-explore metadata -p pattern1 -p pattern2 file1 file2 file3`

nixio/cmd/nixexplore.py
jgrewe: fair enough, avoided assertions

achilleas-k:
Noticed one more minor issue:
The progress printing can interfere with the printed data.

For instance:
```
❯ nixio-explore dump -p "PreContrast" relacs_data.nix
# File: relacs_data.nix
# entity: rectangle-1_PreContrast
# type: relacs.feature.mutable
# id: 130e0c91-5d2c-42a4-937c-e9d26c7b2b56
# created at: 2019-06-18 18:08:41
# last edited at: 2019-06-18 18:08:41

# PreContrast
# None
0.000000   0.000000------------------------------------------] 0.00% ...Dumping ...
1.000000   -0.000752
2.000000   0.001289
3.000000   0.001195
4.000000   0.000748
5.000000   0.001141
6.000000   0.000777
7.000000   0.000692
8.000000   0.000711
9.000000   0.000563
10.000000   0.000472
11.000000   0.001089
12.000000   0.001267
13.000000   0.000540
14.000000   0.000385
15.000000   0.001512
```

It's not an issue if you redirect one of the two (stdout or stderr) but it can cause issues.

nixio/cmd/nixexplore.py
```diff
@@ -0,0 +1,638 @@
...
+    parent_parser.add_argument("-c", "--case_sensitive", action="store_true", help="matching of"
+                               + " entitiy names and types is case sensitive, by default the case is ignored")
+    parent_parser.add_argument("-fm", "--full_match", action="store_true", help="names and types must"
+                               + " be full matches, bey default a partial match is sufficient")
+
+
+def create_metadata_parser(parent_parser):
+    meta_parser = parent_parser.add_parser("metadata", help="Filter and display metadata", aliases=["mdata"],
+                                           description="Search for metadata items or display metadata (sub)trees.")
+    meta_parser.add_argument("-p", "--pattern", type=str, default="", action="append", help=mdata_pattern_help)
```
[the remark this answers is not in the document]
jgrewe: This, I think, can be considered a bug in argparse. I tested this here without a default and it was working fine, then I added the default value before committing. (should have tested again...)

With the last commit I removed the default again and it works as intended. 
```nixio-explore metadata -p "cell type" -p "species" nix_test.h5``` works as expected, at least on my machine.

Regarding the rationale behind this is that I often look for sets of metadata, say, the species and cell type. This should be possible with a single command. In the data parts I would like to eventually support multiple patterns with AND or OR operators.

jgrewe:
@achilleas-k The overlay is indeed kind of annoying, but I am not sure if and how I would solve this, dumping large datasets and redirecting it to file may take a while, thus I want to provide some user feedback. Just dumping it to the command line is actually not the intended way of using dump.
What do you think?

achilleas-k:
I think there's a couple of ways around this.

The command can have an output file option.  When it's specified, the user gets a progress bar printed to stderr (or stdout) while the file is being written.  If it's not specified, the progress bar isn't displayed and the information goes to stdout.

Another solution would be to print the progress bar (to stderr) and not print the output until everything is done.  It can collect the output in a single long string and print it out at the end, separating it from the progress bar.

jgrewe:
decided for the additional ``outfile`` argument. If not given, print is printing to stdout and no progress bar is shown, if given the content goes to file directly with a progress bar.

jgrewe:
closing this one because rebasing it led to hell, replaced by #471

Issue #463, opened 2020-05-08
Inserting odML float property


honzaseda:
Hi, I'm having troubles inserting any odML properties of a float data type into NIX.
I'm running the nixio 1.5.0b4 and the nixodmlconverter version 0.0.7

The problem is with the compatibility check, inserting float values to a NIX property results in: `TypeError: Type 'float' is incompatible with property values`

An example odML property I am trying to insert into NIX:
```
      <property>
        <id>32bb84ad-ee2b-4f60-b001-ee7af098b97b</id>
        <name>Gain</name>
        <value>-1.0</value>
        <type>float</type>
      </property>
```

When debugging, the problem occurs at `nixio.property.OdmlType.compatible`:
```
elif (self == self.Float and
    DataType.get_dtype(value) == DataType.Float):
  return True
```
The loaded OdmlType is float, so the first part of the condition is True, however, for the second condition, the `DataType.get_dtype` method returns `DataType.Double` instead of `DataType.Float`, because of this in the `DataType.get_dtype`:
```
elif isinstance(value, Real):
  return cls.Double
```
This way, the condition for float data types can never be met? Shouldn't the `DataType.get_dtype` return `cls.Float` instead of `cls.Double`?

achilleas-k:
> ```
> elif (self == self.Float and
>     DataType.get_dtype(value) == DataType.Float):
>   return True
> ```
> 
> The loaded OdmlType is float, so the first part of the condition is True, however, for the second condition, the `DataType.get_dtype` method returns `DataType.Double` instead of `DataType.Float`, because of this in the `DataType.get_dtype`:
> 
> ```
> elif isinstance(value, Real):
>   return cls.Double
> ```
> 
> This way, the condition for float data types can never be met? Shouldn't the `DataType.get_dtype` return `cls.Float` instead of `cls.Double`?

Thanks for digging into this.  You're right that the condition would never be true there.  Though the fix would be to check for `DataType.Double` in the compatibility check.

Since Python doesn't distinguish between floats and doubles (it only has double-precision floats that it calls `float`), any float coming from elsewhere is automatically upgraded to (or assumed to be) Double.  In NIXPy we have both DataType.Float and DataType.Double which map to numpy types, which distinguishes between 32 bit floats and doubles.  The `get_dtype()` function is used for mapping native Python types to the NIX (numpy) data types, so it should never return Float,  since `DataType.Float` is `numpy.float32` and this could lead to loss of precision.

[...]

PR #464, opened 2020-05-08
Fixes and simplifications for DataTypes and DataType mappings

achilleas-k:
- Added DataType groups (integers and floats) to simplify data type detection and mapping.
- Small fixes for mapping between odML and NIX property DataTypes.
    - DataType.get_dtype() returns Double for all floating point values (see #463).


achilleas-k:
Rebased

PR #477, opened 2020-10-06
RangeDimension: Replace ticks with link and vice versa when setting

### LINKED ISSUE #476: Replace ticks with link and vice versa
When a RangeDimension has ticks and an object is linked to it, remove the ticks before linking.
Same for the opposite situation: If a RangeDimension has ticks and ticks are added, remove the link.

achilleas-k:
When setting the ticks of a RangeDimension that has an object link (DataArray or DataFrame), the object is unlinked.
When linking a data object to a RangeDimension that has ticks, the ticks are removed.

Closes #476.

### DIFF
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -129,12 +129,14 @@ class DataArray(Entity, DataSet):
         :rtype: RangeDimension
         """
         index = len(self.dimensions) + 1
+
         rdim = RangeDimension.create_new(self, index, ticks)
-        if label:
-            rdim.label = label
-            rdim.unit = unit
+        rdim.label = label
+        rdim.unit = unit
         if self.file.auto_update_timestamps:
             self.force_updated_at()
+        if ticks is not None:
+            rdim.ticks = ticks
         return rdim
 
     def delete_dimensions(self):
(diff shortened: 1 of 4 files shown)


achilleas-k:
Note the message in commit 3307378b793bf8f4fe24d1d4fd59662c9781fb1a

    [h5group] Don't delete empty containers

    The initial consideration was to have "clean" deep objects that didn't
    have lingering empty containers (like a Source container), but this
    becomes a problem when the child object of a non-Container object is
    removed (e.g., RangeDimension.ticks).  The result was the dimension
    object itself was being deleted because it had no children between the
    deletion of the ticks and the creation of the link.  We could reverse
    the operation to get around this, but I think it's better to avoid this
    risky self-deletion in general.

jgrewe (dismissed):
lgtm

jgrewe:
@achilleas-k  instead of not cleaning up the (temporally) empty containers, the method could also take a boolean that indicates whether to remove or not. In that way we could have a default on removing empty containers and leave them alone when needed? What do you think?

achilleas-k:
That's probably a good idea.  It's an internal method so wont affect the API.  I'll add it and update the PR.

achilleas-k:
Done!

PR #478, opened 2020-10-06
Support linking DataFrames to Tags via Features

achilleas-k:
This PR adds support for linking DataFrames to Tags via Features.

One thing I'm unclear about is whether it makes sense for all kinds of links.  Would it make more sense to restrict DataFrame linking to specific kinds of linking?

### DIFF
--- a/nixio/exceptions/exceptions.py
+++ b/nixio/exceptions/exceptions.py
@@ -77,3 +77,10 @@ class DuplicateColumnName(Exception):
     def __init__(self):
         self.message = "Column names for a DataFrame have to be unique."
         super(DuplicateColumnName, self).__init__(self.message)
+
+
+class UnsupportedLinkType(Exception):
+
+    def __init__(self, objtype, linktype):
+        self.message = "LinkType {} is not supported for {}".format(linktype, objtype)
+        super(UnsupportedLinkType, self).__init__(self.message)
--- a/nixio/exceptions/__init__.py
+++ b/nixio/exceptions/__init__.py
@@ -1,11 +1,11 @@
 from .exceptions import (DuplicateName, UninitializedEntity, InvalidUnit,
                          InvalidAttrType, InvalidEntity, OutOfBounds,
                          IncompatibleDimensions, InvalidFile,
-                         DuplicateColumnName)
+                         DuplicateColumnName, UnsupportedLinkType)
 
 __all__ = (
     "DuplicateName", "UninitializedEntity", "InvalidUnit",
     "InvalidAttrType", "InvalidEntity", "OutOfBounds",
     "IncompatibleDimensions", "InvalidFile",
-    "DuplicateColumnName"
+    "DuplicateColumnName", "UnsuportedLinkType"
 )
--- a/nixio/feature.py
+++ b/nixio/feature.py
@@ -65,10 +65,9 @@ class Feature(object):
     def data(self):
         if "data" not in self._h5group:
             raise RuntimeError("Feature.data: Data object not found!")
-        if "target_type" in self._h5group:
+        objtype = self._h5group.get_attr("target_type")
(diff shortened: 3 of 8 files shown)


jgrewe (dismissed):
lgtm

jgrewe:
we additionally need to make sure, that the retrieve feature data method returns the correct data slices but this can, and probably should be part of a different pull request

achilleas-k:
> we additionally need to make sure, that the retrieve feature data method returns the correct data slices but this can, and probably should be part of a different pull request

Right, that's kinda why I asked about the kinds of linking.  I'll see what makes sense and come back here for comments.

achilleas-k:
Switched to draft because this is far from ready.  The LinkContainer that holds the references returns DataArrays.  They way the container is built can only hold one type of object.  I guess this will come up in the C++ NIX library as well.  How do we handle the structure?  Do you intend to have a `features` container in Tags that can hold both DataArrays and DataFrames?  A common superclass?

achilleas-k:
It's ready now.  I added tests for the `feature_data()` method for both Tags and MultiTags with DataFrame features.  I also fixed the object instantiation when getting the `data` member of a Feature.  Features now have a `target_type` attribute that can be either `"DataArray"` or `"DataFrame"` and the data object is created accordingly based on the value of the property.  If the property is missing, it assumes it's a DataArray for backwards compatibility.

Interestingly, no changes were required for the DataView class to be able to index DataFrames.

achilleas-k:
How do we feel about disabling compatibility tests or running them separately with failures allowed so we can actually know when tests pass?

achilleas-k:
Lot of issues cropping up with older tests as well.  Will have to look into it more tomorrow.

achilleas-k:
Update on the errors:
h5py changed some of its string handling in v3 and now `None` is not a valid value for string datasets or attributes.
See https://github.com/h5py/h5py/commit/8d257ffb12e05d21a6ce67832c48f1a8d521cfba

We have a few locations where None is set as a value for unset attributes.  More importantly, the `units` for DataFrame objects are a list of string values, one for each column, and None was a valid unit value for columns that don't have a unit.  This isn't an issue in DataArray since an array without units can just leave the unit attribute unset.

I think the most straightforward solution is to regard empty string as the new "unitless" or None unit.

achilleas-k:
Switched to draft until #485 is reviewed and merged.

achilleas-k:
Rebased and ready

jgrewe (dismissed):

nixio/exceptions/exceptions.py
```diff
@@ -77,3 +77,11 @@ class DuplicateColumnName(Exception):
     def __init__(self):
         self.message = "Column names for a DataFrame have to be unique."
         super(DuplicateColumnName, self).__init__(self.message)
+
+
+class UnsupportedLinkType(Exception):
+
+    def __init__(self, objtype, linktype):
+        self.message = "LinkType {} is not supported for {}".format(objtype,
```
gicmo, line 85: Are the arguments mix up here? Also, what is the minimal python version we are targeting? I have been using `f"LinkType {linktype}"` in other projects and that is so much nicer to read.

achilleas-k: Yeah, want to use f-strings everywhere too, but we're also targeting Python 2 for this release only and dropping it immediately afterwards.  It's weird to still target Py2, but 1.5 was planned for release over a year ago so it stayed in the plan since then.

achilleas-k: Fixed the arg order.

gicmo: Python 2, yo say. ...wow, ok. I am sorry. ;)

gicmo (approved):
:ok_person:

PR #485, opened 2020-11-16
H5Py string handling changes and fixes

achilleas-k:
This PR takes care of [recent changes to string handling in H5Py](https://docs.h5py.org/en/stable/strings.html) as well as some older inconsistencies with string handling.

- Use the h5py string_dtype with explicit UTF-8 encoding, instead of a numpy type.
- Write empty string in place of `None` for DataFrame.units: `None` is no longer a valid element of a string array in h5py.
- Change DataFrame tests to check for bytes instead of strings in returned data.

Copying commit message for DataFrame string/bytes element type tests:
String data and attributes in NIXPy are generally returned as strings.  H5Py returns string data from a file as bytes and these are decoded to strings before returning to the user.  Unfortunately, there's no clean way to do the same with compound data types that contain string elements.  Strings in DataFrame data are returned as bytes and tests are adjusted accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ matrix:
       env: pymajor=3
     - python: "3.7"
       os: linux
-      env: pymajor=3 coverage=1
+      env: pymajor=3 compat=1
       dist: xenial
       sudo: true
     - python: "3.7"
       os: linux
-      env: pymajor=3 nocompat=1
+      env: pymajor=3 coverage=1
       dist: xenial
     - language: generic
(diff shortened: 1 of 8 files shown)


achilleas-k:
Oh right, Python2.  Converting to draft and working on those issues too.

achilleas-k:
I think it's ready but it looks like travis is slow today.  Setting to ready.

jgrewe (approved):
nice, lgtm

Issue #488, opened 2020-11-20
Refactor object creation


achilleas-k:
Entity subclasses override the `_create_new()` method for creating objects and it's a bit ugly and complicated to follow. 
 Would be nice to refactor and simplify.

achilleas-k:
We should enable the checks in `pylintrc` when this is done.

PR #489, opened 2020-11-20
Global pylintrc file and code quality fixes

### LINKED ISSUE #472: Project-wide pylintrc file
We should agree on and create a `pylintrc` file for the project root.

http://pylint-messages.wikidot.com/all-codes lists all the codes and messages supported by pylint.

Pylint supports disabling all messages and selectively enabling individual ones, or we could just disable the most annoying ones.

achilleas-k:
Added a pylintrc file with the error/warning codes listed in #472 and addressed all the issues it produced.

- Some codes were disabled because they require large code restructuring to be addressed (opened issue #488).
- Some common "bad" names were added to the list of good names, like `da` for DataArray and `mt` for MultiTag.

Closes #472

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -80,9 +80,8 @@ class Block(Entity):
         if copy_from:
             if not isinstance(copy_from, MultiTag):
                 raise TypeError("Object to be copied is not a MultiTag")
-            id = self._copy_objects(copy_from, "multi_tags",
-                                    keep_copy_id, name)
-            return self.multi_tags[id]
+            objid = self._copy_objects(copy_from, "multi_tags", keep_copy_id, name)
+            return self.multi_tags[objid]
 
         util.check_entity_name_and_type(name, type_)
         multi_tags = self._h5group.open_group("multi_tags")
@@ -105,7 +104,7 @@ class Block(Entity):
                 extcreated = True
             mtag = MultiTag.create_new(self.file, self, multi_tags,
                                        name, type_, positions)
-        except Exception as e:
+        except Exception as exp:
             msg = "MultiTag Creation Failed"
             if poscreated:
                 del self.data_arrays["{}-positions".format(name)]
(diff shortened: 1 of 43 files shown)


achilleas-k:
Travis macOS builds are still timing out.  Not sure what to do about those.

achilleas-k:
Rebased

nixio/block.py
```diff
@@ -228,9 +227,9 @@ def create_data_array(self, name="", array_type="", dtype=None, shape=None,
         if copy_from:
             if not isinstance(copy_from, DataArray):
                 raise TypeError("Object to be copied is not a DataArray")
-            id = self._copy_objects(copy_from, "data_arrays",
+            objid = self._copy_objects(copy_from, "data_arrays",
                                     keep_copy_id, name)
```
jgrewe, line 231: actually I am not sure, if I really think that objid is a better variable name than id ;)
``jan.set_nitpick = true``
Anyway, indentation is broken here
``jamn.set_nitpick = false``

nixio/dimension_type.py
```diff
@@ -15,4 +15,3 @@ class DimensionType(Enum):
     Sample = "sample"
     Range = "range"
     Set = "set"
-    DataFrame = "data_frame"
```
jgrewe, line 18: good catch

nixio/exceptions/__init__.py
```diff
@@ -7,5 +7,5 @@
     "DuplicateName", "UninitializedEntity", "InvalidUnit",
     "InvalidAttrType", "InvalidEntity", "OutOfBounds",
     "IncompatibleDimensions", "InvalidFile",
-    "DuplicateColumnName", "UnsuportedLinkType"
```
jgrewe, line 10: How did you find that one, impressive ;)

jgrewe (approved):

nixio/block.py
achilleas-k: `id` is a builtin function in Python so I don't like using it for variable or function names.

nixio/exceptions/__init__.py
achilleas-k: I think the linter caught it :)

jgrewe (approved):

PR #490, opened 2020-11-26
DataFrame string columns

achilleas-k:
Changes the way strings are handled in DataFrame columns.

The core change is in H5DataSet, where the `read_data()` method checks for two cases:
- if the dtype of the data coming from disk is `vlen_str_dtype` (`h5py.string_dtype(encoding='utf-8', length=None)`), then it decodes all the data to Python string (from bytes) using `ensure_str()` from the six package (for Python 2 compatibility).
- if the dtype has `fields` (struct/compound dtype), the fields are iterated and any `vlen_str_dtype` columns are decoded like above.

Similarly, if the dtype of a dataset is requested, it returns `DataType.String` (`np.unicode_`).

Moving the string conversion down into the H5DataSet class is moreappropriate:
- The data type conversion is HDF5-specific: We convert the dtype of string objects from disk into Python strings to be returned to the upper layers of the library.  Other (hypothetical, future) implementations would likely have other requirements.
- It avoids needing to make the conversion in all the places where data is read.

Tests have been updated to reflect these changes.

### A note about the commits

There's a bit of flip flopping with the commits.  The conversion was originally written in the DataFrame data retrieval function and then moved to the H5DataSet.  If you want, I can squash the commits so the PR is cleaner.

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -334,11 +334,11 @@ class Block(Entity):
             else:  # if col_names is None
                 if data is not None and type(data[0]) == np.void:
                     col_dtype = data[0].dtype
-                    col_name = list(col_dtype.fields.keys())
+                    col_names = list(col_dtype.fields.keys())
                     raw_dt = col_dtype.fields.values()
                     raw_dt = list(raw_dt)
                     raw_dt_list = [ele[0] for ele in raw_dt]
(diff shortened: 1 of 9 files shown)


achilleas-k:
Rebased and ready

jgrewe (approved):
lgtm

PR #491, opened 2020-11-26
[DataView] Use data object's _read_data()

### LINKED ISSUE #482: DataView: return scaled values if polynom_coefficients
A `DataView` returned by `MultiTag.tagged_data` should return the scaled values if the referenced `DataArray` has `polynom_coefficients`.

achilleas-k:
Instead of reading from the underlying DataSet object's _read_data(), call the linked data object's method.  In the case of a DataArray, this will apply polynomial coefficients and the expansion origin to the data slice before returning.

Test added for retrieving data with coefficients and expansion origin through a MultiTag.

Fixes #482.

### DIFF
--- a/nixio/test/test_multi_tag.py
+++ b/nixio/test/test_multi_tag.py
@@ -481,6 +481,39 @@ class TestMultiTags(unittest.TestCase):
         segtag.extents = wrong_ext
         self.assertRaises(IndexError, segtag.tagged_data, 0, 1)
 
+    def test_multi_tag_data_coefficients(self):
+        sample_iv = 0.001
+        x_data = np.arange(0, 10, sample_iv)
+        y_data = np.sin(2 * np.pi * x_data)
+
+        block = self.block
+        da = block.create_data_array("sin", "data", data=y_data)
+        da.unit = 'V'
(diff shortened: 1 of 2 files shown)


nixio/test/test_multi_tag.py
```diff
@@ -481,6 +481,39 @@ def test_multi_tag_tagged_data(self):
...
+
+        mtag = block.create_multi_tag("sin1", "tag", pos)
+        mtag.extents = ext
+        mtag.units = ['ms']
+        mtag.references.append(da)
+
+        assert np.array_equal(da[:2001], mtag.tagged_data(0, 0)[:])
+
+        da.expansion_origin = 0.89
+        assert np.array_equal(da[:2001], mtag.tagged_data(0, 0)[:])
```
jgrewe, line 515: Does taggedData still work inclusive the bounds? I thought we wanted to change this behaviour to excluding the upper bound...

jgrewe (approved):

nixio/test/test_multi_tag.py
achilleas-k: Different PR :)

PR #495, opened 2020-12-10
Tag retrieval: Inclusive and Exclusive stop rule

### LINKED ISSUE #484: (Multi)Tag extent clarification
I thought the extent in Tags is supposed to be the length of the tagged data.
However, the following code gives me length + 1 data because of https://github.com/G-Node/nixpy/blob/2f4ac17f63b3ffb71e28bcd0d76443b7d6d99568/nixio/multi_tag.py#L164

```python
with nix.File.open('test.nix', 'w') as f:
     b = f.create_block('block', 'block')
     arr_y = b.create_data_array('data', 'data', data=np.arange(10000))
     dim = arr_y.append_range_dimension(np.arange(0,10,0.001))
     dim.unit = 's'
     # Tag a stretch of 1 s
     arr_pos = b.create_data_array('pos', 'pos', data=[0])
     arr_ext = b.create_data_array('ext', 'ext', data=[1])
     mt = b.create_multi_tag('tag', 'tag', positions=arr_pos)
     mt.extents = arr_ext
     mt.references.append(arr_y)
 
     # expecting 999 but getting 1000
     print(mt.tagged_data(0, 'data')[-1])
```

Maybe someone can clarify?

### LINKED ISSUE #494: New Tag data retrieval behaviour (inclusive vs exclusive)
Implement the changes discussed in https://github.com/G-Node/nix/pull/822#issuecomment-715324044

achilleas-k:
This PR implements the slicing rules described in https://github.com/G-Node/nix/blob/master/docs/slicing_and_indices.md.
Slicing is by default excludes the right endpoint now.  The endpoint can be included by specifying `stop_rule=nix.SliceMode.Inclusive` in the `tagged_data()` method of a Tag or MultiTag.

To clarify, the new Exclusive rules work as follows:
- The first element in the DataView returned by `tagged_data` is the value at the last index of each dimension that is **less than or equal to** the Tag's `position`.
- The last element in the DataView returned by  `tagged_data` is the value at the last index of each dimension that is **less than** the Tag's `position+extent`.

Using Inclusive changes the second behaviour of the last element to be the same as for the first element:
- The last element in the DataView returned by  `tagged_data` is the value at the last index of each dimension that is **less than or equal to** the Tag's `position+extent`.

Note that when the value of `position+extent` falls between two dimension ticks, the behaviour is the same for both cases.  For example, if a dimension has integer ticks `[0, 1, 2, 3, 4...]` and `position + extent = 2.2`, then the last index of the DataView will always be `2`.

This is covered in the tests and described by diagrams in comments for future reference (because it always takes me a while to figure out what I'm doing and this will definitely help going forward).

The PR also includes a bit of a rewrite of the code for calculating the slices which simplifies and merges the methods for Tag and MultiTag.

Closes #484.
Closes #494.

### DIFF
--- a/nixio/dimensions.py
+++ b/nixio/dimensions.py
@@ -12,6 +12,7 @@ try:
     from collections.abc import Sequence
 except ImportError:
     from collections import Sequence
+from enum import Enum
 
 import numpy as np
 
@@ -20,6 +21,20 @@ from .dimension_type import DimensionType
 from . import util
 from .container import Container
 from .exceptions import IncompatibleDimensions, OutOfBounds
+
+
+class IndexMode(Enum):
+    """
+    IndexMode is used to define the behaviour of the index_of() methods.
+    """
+    # The ones we need for now: More can be added if we expand the functionality.
+    Less = "less"
+    LessOrEqual = "leq"
+    GreaterOrEqual = "geq"
+
+    # Aliases
+    LEQ = "leq"
+    GEQ = "geq"
 
 
 class DimensionContainer(Container):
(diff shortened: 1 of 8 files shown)


nixio/tag.py
```diff
@@ -113,14 +119,44 @@ def create_feature(self, data, link_type):
...
+        refslice = list()
+        if not self.units:
+            units = [None]*len(data.dimensions)
+        else:
+            units = self.units
+
+        for idx, dim in enumerate(data.dimensions):
+            if idx < len(position):
+                pos = position[idx]
+                start = self._pos_to_idx(pos, units[idx], dim, SliceMode.Inclusive)
```
jgrewe, line 132: I think this is not correct (since it is rather convoluted by nature, I might be wrong):
Consider the case of a `SampledDimension` with `sampling_interval = 1` and `offset = 0`. The Tag's position is 0.5.
This will call `index = dim.index_of(pos * scaling, stop_rule == SliceMode.Inclusive)` and eventually `SampledDimension::index_of(..., leq=True)`. This will return (since no exact match) the index 0. For a slice with start and stop the values returned now range from positions < start to somewhere else. For the end positions the calculations appear correct.

nixio/tag.py
```diff
@@ -113,14 +119,44 @@ def create_feature(self, data, link_type):
...
+            else:
+                # Tag doesn't specify (pos, ext) for all dimensions: will return entire remaining dimensions (0:len)
+                start = 0
+            if extent is None or len(extent) == 0:
+                # no extents: return one element
+                stop = start + 1
+            elif idx < len(extent):
+                ext = extent[idx]
+                stop = self._pos_to_idx(pos+ext, units[idx], dim, stop_rule) + 1
```
jgrewe, line 141: why the +1?

nixio/dimensions.py
```diff
@@ -315,21 +315,36 @@ def position_at(self, index):
...
         return index * sample + offset
 
-    def index_of(self, position):
+    def index_of(self, position, leq=True):
         """
         Returns the index of a certain position in the dimension.
 
         :param position: The position.
+        :param leq: Less or Equal mode (default True).
+                    Whether to return the index if it matches the position exactly.
```
jgrewe, line 324: not sure whether I really understand this first sentence. Maybe remove it? 
Not sure, if the `leq` flag alone is sufficient. For starting points of ranges we need the greater of equal.

jgrewe:
I agree with the last position handling, but I am not sure about the start position handling. As I see it, we can only define whether it is less  or less or equal. For numpy like slicing we should have greater or equal for the start position. Or am I mistaken?

nixio/dimensions.py
```diff
@@ -315,21 +315,36 @@ def position_at(self, index):
...
         offset = self.offset if self.offset else 0
         sample = self.sampling_interval
-        index = np.round((position - offset) / sample)
-        if index < 0:
+        scaled_position = (position - offset) / sample
+        if scaled_position < 0:
             raise IndexError("Position is out of bounds of this dimension!")
-        return int(index)
+
+        index = int(scaled_position)
```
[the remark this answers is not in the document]
achilleas-k: According tot he docs:
> Conversion from floating point to integer may round or truncate as in C; see functions math.floor() and math.ceil() for well-defined conversions.

So you're right, it's best to be explicit.

nixio/tag.py
achilleas-k: Yes it will return index 0.  Isn't that correct?

nixio/tag.py
achilleas-k: The `stop` element of a slice isn't inclusive ever.  `Slice(0, 30)` is equivalent to `[0:30]` which returns 30 elements, the last one being index 29.  `_pos_to_idx()` on the `pos+ext` value returns the `last_index`, the index of the last value that should be included in the return slice, so the slice stop value should be `last_index + 1`.

nixio/dimensions.py
achilleas-k: Oh, I read https://github.com/G-Node/nix/blob/master/docs/slicing_and_indices.md#rangematching-enumeration wrong.  Disregard the reply to the first comment then.  You're right, start positions should be `geq` and I made them `leq`.

achilleas-k:
> I agree with the last position handling, but I am not sure about the start position handling. As I see it, we can only define whether it is less or less or equal. For numpy like slicing we should have greater or equal for the start position. Or am I mistaken?

You're not wrong.  I read the document a few times and still did it wrong.  I made the start position follow `leq` rules as well.

nixio/tag.py
achilleas-k: > Yes it will return index 0. Isn't that correct?

Right, I read the doc wrong.  Fixing now.

achilleas-k:
Added an `index_of()` method for SetDimension too.  All the `index_of()` methods take an IndexMode argument now which can be Less, LEQ, or GEQ, and the method does rounding and bounds checking accordingly.

Added tests to the dimensions for all of these.

nixio/tag.py
jgrewe: got it, thanks

jgrewe (approved):
nice, thank you!

Issue #498, opened 2021-05-19
Broken references for nested Source


DanBenHa:
I want to represent data coming from different channels from one probe using nested `Source`s.
For example, if I had two probes with 32 channels each, I would have two `Sources` at the block level. Each of the `Source`s would have unique channel numbers. Concretely, I want to
1. Get the probe identity (block-level `Source`) from referring objects like `DataArray`.
2. Get all referring objects from a probe, e.g. all channels belonging to "region0"

Somehow, this doesn't work:

```
f =  nixio.File.open("test.nix", "w")
b = f.create_block("block0", "session")
s0 = b.create_source("region0", "region")
s00 = s0.create_source("chan0", "chan")
s01 = s0.create_source("chan1", "chan")
da0 = b.create_data_array("da0", "foodata", data=np.zeros(10))
da0.sources.append(s00)
da1 = b.create_data_array("da1", "foodata", data=np.zeros(10))
da1.sources.append(s01)

# trying to find data arrays from "chan0"
print(s00.referring_objects)
# This works
s_from_da = da0.sources[0]
print(s_from_da.referring_objects)
# parent should be "region0", but it's the block
print(s_from_da._parent)
# "s_from_da" is not its parent...
print(s_from_da == s00)
# get all data_arrays from "region0"
print(s0.referring_objects)
```

Did I misunderstand the concept of `Source`?

Edit:
I guess that is related to the note at https://github.com/G-Node/nixpy/blob/fa561104305c73b233d9903d601440c390e2c042/nixio/source_link_container.py#L15
Doesn't sound like a solution is around the corner...
So quickfix would be to have all those channel `Source`s at the top level and put them in a `Group` according to the probe?

Edit 2:
Hmm, apparently `Group` is not considered a `referring_object`...
Resorting to defining the probe/region in the `Source`'s `definition`. Oh dear 😅

jgrewe:
@DanBenHa sorry for the inconvenience. We will have a closer look at it. Your understanding of the source is correct and I think this can be considered a bug. 
When searching for the referring objects we do not check whether the parent is a Block, or a parent Source. In your case of nested sources we need to traverse to the root of the tree.

jgrewe:
fixed with #499 closing this issue

PR #499, opened 2021-05-20
quick fix for parent block access  on sources

jgrewe:
this pr addresses issue #498. 
* adds a ``parent_block`` property to the source which is needed to finding referring objects
* adds an explicit ``parent_source`` property (which at the moment performs a search on the sources tree)

* it **does not** change the behaviour of the ``referring_objects`` methods. They will still only return those objects that link directly to the source. That is, calling ``referring_data_arrays`` on a parent source will so far not collect the *DataArrays* of the child sources. I would like to leave this open for discussion. Since these are properties, there is no way of overloading with a default argument. For the time being, this needs to be done on the user side. We may want to discuss this at some point.

### DIFF
--- a/nixio/source.py
+++ b/nixio/source.py
@@ -48,9 +48,15 @@ class Source(Entity):
             raise exceptions.DuplicateName("create_source")
         src = Source.create_new(self.file, self, sources, name, type_)
         return src
-    
+
     @property
     def parent_block(self):
+        """
+        Returns the block this source is contained in.
+
+        Returns:
+            block [nix.Block]: the Block containing this source
+        """
         if self._parent_block is not None:
             return self._parent_block
         maybe_block = self._parent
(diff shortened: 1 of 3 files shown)


nixio/source.py
```diff
@@ -48,6 +49,51 @@ def create_source(self, name, type_):
...
+    @staticmethod
+    def _find_parent_recursive(source, name_or_id_of_child):
+        if name_or_id_of_child in source.sources:
+            return source
+        else:
+            for s in source.sources:
+                s = Source._find_parent_recursive(s, name_or_id_of_child)
+                if s is not None:
+                    return s
+        return None
```
achilleas-k, line 77: Could this be an instance method with the first arg as `self` instead of a `source`?  Or is the source not always an instantiated NIX object and could just be an H5 reference?

Also, it might be safer to only use IDs for searching and not names.  It's possible for multiple sources to have the same name if they're not on the same level and this could find the wrong parent.
I see the actual implementation below does use the ID, so maybe the argument/variable name could reflect this to make it clear and maybe check if it's a UUID.

achilleas-k (approved):
Looks great!  Thanks Jan.

Tiny nitpick :)

nixio/source.py
jgrewe: @achilleas-k you are right. Please check the current version and regardless of what GitHub claims, of course I did not dismiss your review ;)

achilleas-k: haha
yeah I think it dismisses automatically when the code changes.

achilleas-k (approved):
Perfect.  Thank you :)

PR #503, opened 2021-06-04
Range dim

jgrewe:
fixes
* missing return in ``DimensionLink.id``
* a few docstrings
* the version number in appveyor.yml
* ``SampledDimension.axis`` returned the wrong number of data points (off by 1 due to rounding issues under certain conditions)

adds
* ``RangeDimension.is_alias`` method for downward compatibility
* ``DataArray.append_range_dimension_using_self`` as a convenience method

### DIFF
--- a/nixio/dimensions.py
+++ b/nixio/dimensions.py
@@ -86,7 +86,7 @@ class DimensionLink(object):
 
     @property
     def id(self):
-        self._h5group.get_attr("entity_id")
+        return self._h5group.get_attr("entity_id")
 
     @property
     def file(self):
--- a/nixio/dimensions.py
+++ b/nixio/dimensions.py
@@ -236,16 +236,8 @@ class Dimension(object):
     def index(self):
         return self.dim_index
 
-    def link_data_array(self, data_array, index):
-        if len(data_array.data_extent) != len(index):
-            raise IncompatibleDimensions(
-                "Length of linked DataArray indices ({}) does not match"
-                "number of DataArray dimensions ({})".format(
-                    len(data_array.data_extent), len(index)
-                ),
-                "Dimension.link_data_array"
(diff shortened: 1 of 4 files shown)


jgrewe:
@achilleas-k guess this is ready for review ;)

nixio/dimensions.py
```diff
@@ -466,8 +478,44 @@ def create_new(cls, data_array, index, ticks):
...
+        :returns: True, if this dimension links to a DataArray, False otherwise
+        :rtype: bool
+        """
+        if self._h5group.has_data("ticks"):
+            return False
+        elif not self.has_link and len(self._h5group) > 0:
+            return True
+        elif self.has_link and self.dimension_link._data_object_type == "DataArray":
+            return True
+        return False
```
achilleas-k, line 498: Does alias dimension imply it's a self-referencing dimension?  Should this be checking if the linked DataArray is the same as the DataArray that owns the dimension?

achilleas-k (approved):
Look good.  Approved, though I have one question I'm not sure about.

nixio/dimensions.py
jgrewe: Yes, thought about checking for it but then thought that it actually does not matter for the idea ...
What do you think? Better to keep it strictly in the "old" sense of the aliasRangeDimension?

achilleas-k: I guess it depends what we want "alias" to mean.  It used to be self reference only but we could say now it means "link to a DA".  I have no strong feelings either way.

jgrewe: Since this will die with the use of the AliasRangeDimension, I would say there's no harm done and would argue for keeping it simple..., i.e. as it is now :)

PR #504, opened 2021-06-13
Adds SampledDimension axis overload

jgrewe:
with this it is possible to specifiy a start position, not only a start index

start index takes precedence over start position to maintain old behaviour. added some value checking too.


jgrewe:
closing and reopening this to see what happens to appveyor...

PR #505, opened 2021-06-17
[block] add label and unit as optional args for array creation

jgrewe:


achilleas-k (dismissed):
These constructors are getting a bit unwieldy, but I suppose it's all optional so it's fine.

jgrewe:
@achilleas-k  when writing the example files for the python docs, the two lines to add unit and label were among the most annoying... 
Closing and reopening to trigger appveyor

PR #508, opened 2021-06-20
[SampledDimension] add axis overload to specifiy a start position ...

jgrewe:
not only a start index


achilleas-k (approved):
Looks good.  I considered whether it should be a warning or even an error if both arguments are specified, to make it clear that they don't work together, but maybe keeping the behaviour documented alone should be enough.

PR #516, opened 2021-07-11
Allow DataViews to become invalid fixes #502

### LINKED ISSUE #502: MultiTag.tagged_data: OOB and non-intuitive slicing
Suppose I have a sparsely firing neuron and I want to count how many spikes I have per trial. I'd define a MultiTag as the trials with some positions and extents. With tagged_data I would get all and only those spikes present in a range defined by a position and its extent.
When I try this, I get out-of-bound errors or incorrect data. Wouldn't it be possible to just return an empty numpy array if there is no data within a range? This is probably a design choice but it seems a bit weird to me.

```
timestamps = np.array([2.8, 3.8])

f = nixio.File.open("test.nix", "w")
b = f.create_block("block0", "foo")

da_ts = b.create_data_array("spike_times", "foo", data=timestamps)
da_ts.unit = "s" 
da_ts.label = "time"
dim = da_ts.append_range_dimension()
dim.link_data_array(da_ts, [-1])

# only trial 2 (3.8 to 4.2) has a spike
pos = b.create_data_array("pos", "foo", data=[2.0, 3.0, 3.8, 6.4])
ext = b.create_data_array("ext", "foo", data=[0.4, 0.4, 0.4, 0.4])
mt = b.create_multi_tag("trials", "foo", positions=pos)
mt.extents = ext 
mt.references.append(da_ts)
# trial 0: no spikes: expect empty array, get OOB
print(mt.tagged_data(0, "spike_times")[:])
# trial 1: no spikes: expect empty array (or at least the closest spike), get next trial's spike
print(mt.tagged_data(1, "spike_times")[:])
# trial 2: 1 spike: expect 1 spike, get the correct spike
print(mt.tagged_data(2, "spike_times")[:])
# trial 3: no spikes: expect empty array, get OOB
print(mt.tagged_data(3, "spike_times")[:])
```

jgrewe:
This PR changes the way ranges along a dimension are calculated. 
* allows DataView to be invalid and reading access will return an empty array
* range calculation is delegated to the dimensions respecting the SliceMode.
* allows for some simplification of the BaseTag._calc_data_slice method
* We do not necessarily return at minimum a single value along a given dimension. If there are no data points between start and end position the range is invalid and the DataView will become invalid and return an empty array. Just like slicing a numpy array that does not return a result. 
* Invalid DataViews return no data extent, write access will raise an exception. They offer a "debug message" containing the reason of their invalidity.

* I might be that I went a little far with removing the OOB and IncompatibleDimensions exceptions from DataView, happy to have your opinions..

### DIFF
--- a/nixio/tag.py
+++ b/nixio/tag.py
@@ -122,7 +122,6 @@ class BaseTag(Entity):
             units = self.units
 
         for idx, dim in enumerate(data.dimensions):
-            scaling = 1.0
             if idx < len(position):
                 start_pos = position[idx]
                 start_pos, scaling = self._scale_position(start_pos, units[idx], dim)
--- a/nixio/tag.py
+++ b/nixio/tag.py
@@ -122,30 +122,31 @@ class BaseTag(Entity):
             units = self.units
 
         for idx, dim in enumerate(data.dimensions):
+            scaling = 1.0
             if idx < len(position):
-                pos = position[idx]
-                start = self._pos_to_idx(pos, units[idx], dim, IndexMode.GreaterOrEqual)
-            else:
-                # Tag doesn't specify (pos, ext) for all dimensions: will return entire remaining dimensions (0:len)
-                start = 0
-            if extent is None or len(extent) == 0:
-                # no extents: return one element
-                stop = start + 1
-            elif idx < len(extent):
-                ext = extent[idx]
-                stop = self._pos_to_idx(pos + ext, units[idx], dim, stop_rule.to_index_mode()) + 1
-            else:
-                # Tag doesn't specify (pos, ext) for all dimensions: will return entire remaining dimensions (0:len)
-                stop = data.shape[idx]
+                start_pos = position[idx]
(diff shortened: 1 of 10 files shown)


mpsonntag (approved):
Looks good to me

nixio/tag.py
```diff
@@ -135,33 +123,64 @@ def _calc_data_slices(self, data, position, extent, stop_rule):
...
+        if dimtype == DimensionType.Set:
+            if unit and unit != "none":
+                raise IncompatibleDimensions(
+                    "Cannot apply a position with unit to a SetDimension",
+                    "Tag._pos_to_idx"
+                )
+        else:
+            if dimunit is None and unit is not None:
+                raise IncompatibleDimensions(
+                    "Units of position and SampledDimension "
```
mpsonntag, line 170: Should "RangeDimension" also be considered at this point?

mpsonntag (approved):
Looks good to me now.

PR #517, opened 2021-07-13
Add documentation examples tests

mpsonntag:
This PR introduces tests for the documentation examples in the `docs` folder. Since `docs` resides outside the nixio package, the testing requires executing the example files directly, copying the required 'lenna.png' image from the docs folder and cleaning up created test nix files once the tests are done. Running and cleanup works, please comment if there are any reservations to the way these tests are run or if individual tests should not be excluded.

Running tests for these examples requires the additional packages `matplotlib, scipy, pillow` to be installed. These packages have been added to the `setup.py: tests_requires` and the CI config files.

All plots are rendered non-blocking and are closed after the tests are done. The examples "imageData.py" and "singleROI.py" will open an image via `pillow image.show()` which cannot be closed automatically. These tests block on Windows and are only run on Linux OS.

The PR also fixes
- invalid old style property values usage in the `imageWithMetadata` example.
- all usages of the deprecated `DataArray.data` method.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -37,7 +37,7 @@ jobs:
         pylint --rcfile=.pylintrc nixio
     - name: Test with pytest
       run: |
-        pip install pytest
+        pip install pytest scipy pillow
         # pytest will skip cross-compatibility tests since NIX isn't installed
         pytest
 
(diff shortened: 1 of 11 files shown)


jgrewe (approved):
I like it, one thing for later might be to run these tests only once

PR #520, opened 2021-07-27
Allow for list-like labels of SetDimension

JuliaSprenger:
Currently `append_set_dimension` fails if a non-list object, e.g. numpy array, is provided as labels. For convenience it would be nice to accept any object that can be converted to a list. The same is in principle true for Dimension ticks, but here no check on the input type is currently performed.

### DIFF
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -91,7 +91,7 @@ class DataArray(Entity, DataSet):
         index = len(self.dimensions) + 1
         setdim = SetDimension.create_new(self, index)
         if labels is not None:
-            if not isinstance(labels, list):
+            if not isinstance(labels, (list, str)):
                 labels = list(labels)
             setdim.labels = labels
         if self.file.auto_update_timestamps:
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -91,10 +91,11 @@ class DataArray(Entity, DataSet):
         index = len(self.dimensions) + 1
         setdim = SetDimension.create_new(self, index)
         if labels is not None:
-            if not hasattr(labels, '__iter__'):
-                labels = str(labels)
-            if isinstance(labels, str):
-                labels = [labels]
+            if not hasattr(labels, '__iter__') or isinstance(labels, str):
+                raise ValueError('`labels` has to be a list-like object.')
+            for label in labels:
+                if not isinstance(label, str):
+                    raise ValueError(f'`labels` has to contain string objects, not {type(label)}')
             if not isinstance(labels, list):
                 labels = list(labels)
             setdim.labels = labels
--- a/nixio/data_array.py
+++ b/nixio/data_array.py
@@ -82,6 +82,9 @@ class DataArray(Entity, DataSet):
         Append a new SetDimension to the list of existing dimension
         descriptors.
 
+        :param labels: The set of sample labels
+        :type labels: list or convertible to list
+
         :returns: The newly created SetDimension.
         :rtype: nixio.SetDimension
         """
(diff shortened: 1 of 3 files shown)


jgrewe:
I generally like it. Only problem I see is if the user passes a string
```
labels = "string"
print(list(labels))
['s', 't', 'r', 'i', 'n', 'g']
```

JuliaSprenger:
Ah, yes. String need to be treated differently again. How about only converting if it's not a string? Or do you know more exceptions like this?

jgrewe:
I think string is the only exception, do you agree @achilleas-k ?

JuliaSprenger:
On a second thought, since labels should match in principle the length of the set, providing a single string is not a valid input anyway. So it should either result in an error or should really be interpreted as single character labels, right?

jgrewe:
not necessarily, you could have a dimension with a single entry and want to attach a label, that would be a simple string.

Actually the following will fail at the moment:
```
da = block.create_data_array("test", "test", data=np.ones((1, 30)))
da.append_set_dimension(labels="label") 
```
this will work:
```
da.append_set_dimension(labels=["label"]) 
```
I guess we need to handle strings differently. if the input is a string, will need to make it a list with a single entry. `labels=[labels]` Whether or not the number of labels matches the size of the data cannot be tested at this point since the DataArray may grow in size.

achilleas-k:
> I guess we need to handle strings differently. if the input is a string, will need to make it a list with a single entry. `labels=[labels]`

Not sure I like this.  I think it's good that it fails with a single string even if it's a single dimension.  `labels` implies a collection of multiple labels.  Sure a single dimension only has one, but programmatically it should still be wrapped in a collection with `n=1` imo.

jgrewe:
fair enough but the let it fail with a nixio exception and not the current h5py exception.
```
/usr/local/lib/python3.8/dist-packages/h5py/_hl/filters.py in fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts, shuffle, fletcher32, maxshape, scaleoffset, external)
    114         if any((chunks, compression, compression_opts, shuffle, fletcher32,
    115                 scaleoffset is not None)):
--> 116             raise TypeError("Scalar datasets don't support chunk/filter options")
    117         if maxshape and maxshape != ():
    118             raise TypeError("Scalar datasets cannot be extended")
```
sorry @JuliaSprenger  for the mess

achilleas-k:
Yes, we should catch it instead and produce a meaningful error message

JuliaSprenger:
Ok, so how about the latest version? Or do you think the checking of each item might be too compute intense for real datasets?

nixio/data_array.py
```diff
@@ -82,12 +82,22 @@ def append_set_dimension(self, labels=None):
...
         setdim = SetDimension.create_new(self, index)
-        if labels:
+        if labels is not None:
+            if not hasattr(labels, '__iter__') or isinstance(labels, str):
+                raise ValueError('`labels` has to be a list-like object.')
+            for label in labels:
+                if not isinstance(label, str):
+                    raise ValueError(f'`labels` has to contain string objects, not {type(label)}')
+            if not isinstance(labels, list):
+                labels = list(labels)
```
achilleas-k, line 100: I think it would make more sense to move these checks to the `labels` property setter in the `SetDimension` so that the same checks are applied when doing `set_dim.labels = ...`

achilleas-k:
> Ok, so how about the latest version? Or do you think the checking of each item might be too compute intense for real datasets?

I don't think it's a big issue.  Probably fine.

mpsonntag (approved):
Looks good to me now.

PR #521, opened 2021-08-08
[RangeDim] fix downward reading issue with labels and units

jgrewe:


mpsonntag (approved):
LGTM, but if this is a fix, would it make sense to also immediately add a test with this PR?

jgrewe:
Hi @mpsonntag you are right, bad thing is that we would need to create a version 1.1 file to test it.
Would rather delay this topic.

mpsonntag:
@jgrewe, ok, then I think we can merge.

PR #522, opened 2021-08-08
Set dim label

jgrewe:
as we discussed, this pr adds a label for SetDimensions 

As this is a non-breaking change to the model the file format version is increased to 1.2.1


mpsonntag (approved):
Looks good to me.

PR #523, opened 2021-08-11
[pprint] max_depth=-1 is now default

jgrewe:


mpsonntag (dismissed):
LGTM

achilleas-k (dismissed):

achilleas-k:
Needs a rebase.

jgrewe:
@achilleas-k  thanks for pinging... Will do but I also need to (re)consider the side effect of this choice on the nixio command line tool, which will then print out the whole metadata tree.
Is that a bad choice? I am not sure. If yes, then I will limit the default recursion depth there. 
What do you think?

achilleas-k:
I  think it's fine to spew everything out by default.  It might be a bit annoying, but no more than any other command line tool that produces multiple pages of output by default.

jgrewe:
@achilleas-k  rebased ;)

gicmo (approved):
LGTM
(naughty whitespace changes ;)

PR #529, opened 2021-10-21
Data array slice

jgrewe:
ran into a bug with get slice and range dimensions...


mpsonntag (dismissed):
Looks good to me.

mpsonntag (approved):
Looks still good to me.

PR #531, opened 2021-10-25
[h5dataset] change string conversion to work with more than 1d

jgrewe:
fixes a problem with 2d string data


mpsonntag (dismissed):
Looks good!

I have a general comment though: if it was an issue and all tests went through, I assume the tests did not cover this so far. Would it make sense to add a test, that covers the 1d / 2d / nd cases with this PR?

mpsonntag (approved):
Looks good to me

PR #532, opened 2021-10-26
addresses a few issues

### LINKED ISSUE #465: drop support for old python versions
we should drop support for python 2 and probably below 3.6 since the upgrade tool contains f-string handling that was introduced with 3.6

### LINKED ISSUE #524: Creating a block with only "name" provided ignores the "name" value
When creating a block and providing only the name, but not the type, the name value is ignored and the block name defaults to the blocks ID.

e.g. 
`b = nf.create_block(name="I-shall-be-ignored")`

The corresponding lines responsible can be found in `block.py` and `entity.py`: 

https://github.com/G-Node/nixpy/blob/c5ffc3b6b6d22b89433acffc9b96a6a0dacfe3e0/nixio/block.py#L55
https://github.com/G-Node/nixpy/blob/c5ffc3b6b6d22b89433acffc9b96a6a0dacfe3e0/nixio/entity.py#L23

Is this expected behavior and what would be the reason for ignoring a name value when the type has not been provided?

### LINKED ISSUE #525: [Block] Inconsistent ValueError/DuplicateName exception
In NIX, creating any entity (`DataArray`, `Tag`, `Section`, etc) with the identical name of an existing entity on the same level will lead to a `DuplicateName` exception.

Except for the `Block` entity, where a duplicate name entry leads to a `ValueError`.

Is there a specific reason for the differing behavior?

jgrewe:
fixes #524 
closes #525 
closes #465

### DIFF
--- a/nixio/entity.py
+++ b/nixio/entity.py
@@ -22,11 +22,10 @@ class Entity(object):
     def create_new(cls, nixfile, nixparent, h5parent, name=None, type_=None):
         if name and type_:
             id_ = util.create_id()
-            util.check_entity_name_and_type(name, type_)
         if not name:
             name = util.create_id()
             id_ = name
-            util.check_entity_type(type_)
+        util.check_entity_name_and_type(name, type_)
 
         h5group = h5parent.open_group(name)
         h5group.set_attr("name", name)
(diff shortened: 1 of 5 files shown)


mpsonntag (dismissed):
These are very lovely fixes, thanks!

mpsonntag (dismissed):
Very nice fixes!

mpsonntag (approved):

PR #533, opened 2021-12-04
[dorelease] add function to find a valid python command

jgrewe:
in my installation the default ``python`` still maps to python 2.7 ... I need to call python3

### DIFF
--- a/scripts/dorelease.py
+++ b/scripts/dorelease.py
@@ -251,8 +251,9 @@ def check_python_command():
         ok = call([cmd, "--version"])
         if ok == 0:
             return cmd
-    except:
-        return None
+    except FileNotFoundError:
+        cmd = None
+    return cmd
 
 
 def main():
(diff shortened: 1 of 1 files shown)


mpsonntag (approved):
Looks good to me

scripts/dorelease.py
```diff
@@ -241,6 +241,21 @@ def print_appveyor_urls():
...
+    if " 3." in py_version:
+        return cmd
+    cmd = "python3"
+    try:
+        ok = call([cmd, "--version"])
+        if ok == 0:
+            return cmd
+    except FileNotFoundError:
+        cmd = None
+    return cmd
```
achilleas-k, line 256: Since we don't support python2 at all anymore, is this necessary?  We can just call `setup.py` with `python3` in `main` and if it's not available it wont work.

jgrewe: You're right if  there will always be a python3 command even when no python2 is installed. Is that the case?

achilleas-k: Yes, `python 3.x` (any version) should always be available as `python3`.

PR #537, opened 2022-03-12
Invalid event slices

jgrewe:
With event data (RangeDim DataArrays) it can easily happen, that there are no events in a certain time frame. In the current version trying to read such a slice will lead to a ``ValueError: length should not be negative``. 

This pull request catches such situations and returns an empty slice.


mpsonntag (dismissed):
Looks good to me

mpsonntag (approved):
GH actions crisis averted, LGTM

PR #538, opened 2024-02-17
test_doc_examples.py: initialize examples_path.

emollier:
While trying to package the Python nixio module in Debian, most tests from test_doc_examples.py are failing with the following error:

	    def handle_lif(self):
	>       lif_path = Path.joinpath(self.examples_path, "lif.py")
	E       AttributeError: 'TestDocumentationExamples' object has no attribute 'examples_path'

This looks to be caused by missing initialisation of the examples_path attribute of the TestDocumentationExamples class.  Initialising the attribute with the path to the documentation from the root of the sources corrects the issue.

### DIFF
--- a/nixio/test/test_doc_examples.py
+++ b/nixio/test/test_doc_examples.py
@@ -15,6 +15,8 @@ TEST_IMAGE = "lenna.png"
 
 class TestDocumentationExamples(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


jgrewe (approved):

jgrewe:
Thank you @emollier for packaging and fixing!

PR #541, opened 2024-04-28
remove Python2 crumbs

a-detiste:
Hi !

I saw this new package in Debian today popping up in my QA list.

Here's a cleanup of Python2 code.

@emollier : please remove python3-six dependency when a new release is published, or I'll do if I notice first

https://tracker.debian.org/pkg/python-nixio

https://salsa.debian.org/detiste-guest/python3-six-removal/-/commit/b3218f61adb31c09cac2de735fd83997abe9b707

### DIFF
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -15,7 +15,6 @@ try:
     from collections.abc import OrderedDict
 except ImportError:
     from collections import OrderedDict
-import sys
 
 from .util import find as finders
 from .compression import Compression
@@ -299,13 +298,6 @@ class Block(Entity):
             return self.data_frames[objid]
 
         util.check_entity_name_and_type(name, type_)
-        if (isinstance(col_dict, dict)
-                and not isinstance(col_dict, OrderedDict)
-                and sys.version_info[0] < 3):
-            raise TypeError("Cannot create a DataFrame from a dictionary "
-                            "in Python 2 as the order of keys is not "
-                            "preserved. Please use the OrderedDict class "
-                            "from the collections module instead.")
 
         if data is not None:
             shape = len(data)
--- a/nixio/block.py
+++ b/nixio/block.py
@@ -7,10 +7,8 @@
 # modification, are permitted under the terms of the BSD License. See
 # LICENSE file in the root of the Project.
 
-try:
-    from sys import maxint
-except ImportError:
-    from sys import maxsize as maxint
+from sys import maxsize
+
 import numpy as np
 from inspect import isclass
 try:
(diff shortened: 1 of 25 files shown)


emollier:
Hi Alexandre,

Thanks for your QA work, I'll try to not forget about the python3-six package once removed upstream; I left a Fixme item in the d/control file so lintian will ring some alarm bells for me when the time will come.

Have a nice day,  :)
Étienne.

a-detiste:
I discoverd this feature :-) 

image

But messing with UDD & postgres is fun too

jgrewe (approved):
lgtm, thanks!

nixio/block.py
```diff
@@ -358,7 +349,7 @@ def create_data_frame(self, name="", type_="", col_dict=None,
         if col_dict is not None:
             for nam, dt in col_dict.items():
                 if isclass(dt):
-                    if any(issubclass(dt, st) for st in string_types) \
+                    if any(issubclass(dt, st) for st in str) \
```
achilleas-k, line 352: A lot of the failures seems to be related to these changes, where an iterable `string_types` was replaced with a plain type `str` and the iteration wasn't dropped.
This should change to:
```
if issubclass(dt, str):
```
(or `isinstaince()`) since we only have one string type now.

achilleas-k (changes requested):
Thanks for bringing this to our attention and doing a lot of the work to clean it up.  There appear to be a few issues with iterating on a single type (see inline comment).

I'd also love to see this PR split into multiple commits if it's not too much trouble.  For example, the changes in `setup.py` could be their own commit.

I understand if you don't have time to work on this and perhaps this was meant as a quick patch to make the project buildable for the distro and bring it up to date a bit, so let us know and we can take it from here if necessary.

nixio/block.py
a-detiste: Hi, I don't mind

a-detiste: If I'm not failing I'm not trying hard enough. Here the codebase was a tiny bit confusing because in some places it is know that `instance()` will work on a tuple and in others places there was this `any()` construct.

a-detiste:
I'm out for the day :-) You can merge as-is if you like it of course. `six.ensure_str()` & `six.ensure_text()` are always bit more complicated to replace.

a-detiste:
I on Holliday for one week. Do you still need my help before merging this ?

jgrewe:
Hi @a-detiste, just enjoy your holidays :) 
Let me poke @achilleas-k to draw his attention to this pr again.
Thanks again.

achilleas-k (approved):

PR #542, opened 2024-07-08
Numpy

jgrewe:
With numpy 2.0 we run into several issues. One is solved here, but there are more changes regarding string/unicode handling that kill the continuous integration at the moment. Most of them I could not reproduce locally. As an intermediate solution I limited numpy to versions less than 2.0.

This pr fixes:
* miniconda action by specifying the version to "latest"
* switches to ``pathlib`` instead of ``os`` in test_docs
* load_module deprecation

@achilleas-k if you could spare a quick look ;)


achilleas-k (approved):
Thanks!  lgtm

PR #544, opened 2024-11-10
finish six removal

a-detiste:
Hi,

This is follow-up to https://github.com/G-Node/nixpy/pull/541 I forgot to do.

@emollier

### DIFF
--- a/scripts/dorelease.py
+++ b/scripts/dorelease.py
@@ -105,7 +105,7 @@ def update_info(newver):
         infodict = json.load(infofile)
 
     verstring = infodict["VERSION"]
-    newverstring = re.sub("[0-9\.a-z]+", newver, verstring)
+    newverstring = re.sub(r"[0-9\.a-z]+", newver, verstring)
 
     if newverstring == verstring:
         print("No changes required in info.json")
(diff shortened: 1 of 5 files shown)


scripts/dorelease.py
```diff
@@ -105,7 +105,7 @@ def update_info(newver):
         infodict = json.load(infofile)
 
     verstring = infodict["VERSION"]
-    newverstring = re.sub("[0-9\.a-z]+", newver, verstring)
+    newverstring = re.sub(r"[0-9\.a-z]+", newver, verstring)
```
emollier, line 108: Sligthly unrelated, but useful nonetheless, these r"" raw strings will fix a bunch of SyntaxWarnings which would stem to the failure to translate the escape code "\\." in this particular case.

a-detiste: Yes, I always run `pyflakes` before submitting a PR; and the warnings were there, so ...

emollier: Acknowledged, I guess that explains it and the programming languages selection in setup.py.  :)

emollier:
Thanks Alexandre for the follow-up changes to remove traces of six!

I went through your changes and I believe they are sound, but I am only a passer-by contributor to nixio/nixpy, so better wait for main upstream authors to comment.

Have a nice day,  :)
Étienne.

achilleas-k (approved):
LGTM.  `ensure_str()` could be a utility function so it's not duplicated but that's minor and we can do it in a separate PR.

a-detiste:
We ll ... ensure_str() was there in the first time because the Py2 was not sure what is was getting. This could be removed after carefully auditing wether these are internal API with safe input or API that might get unexpected data from callers.

This kind of audit can be greatly aided by adding type annotations all over the place and running `mypy`.

But many people feel that type annotations doesn't belong in Python.

I may try to do the bare minimum needed to  remove the two `ensure_str` If you are interested.

Greetings

a-detiste:
`data_to_value` is a copy of `ensure_str` with a different name.

Issue #551, opened 2025-03-24
Cannot install latest version of nixio


drodarie:
I am not sure about what is causing the issue but I got the following error when trying to install nixio in a python 3.10 environment:

```bash
pip install nixio
```
```
Requirement already satisfied: nixio in ./workspace/venv/lib/python3.10/site-packages (1.5.3)
Collecting nixio
  Downloading nixio-1.5.4.tar.gz (28.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 28.8/28.8 MB 12.4 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [27 lines of output]
      Traceback (most recent call last):
        File "/home/drodarie/workspace/venv/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
        File "/home/drodarie/workspace/venv/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
        File "/home/drodarie/workspace/venv/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 143, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/tmp/pip-build-env-eydu5ae2/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 334, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
        File "/tmp/pip-build-env-eydu5ae2/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 304, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-eydu5ae2/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 522, in run_setup
          super().run_setup(setup_script=setup_script)
        File "/tmp/pip-build-env-eydu5ae2/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 320, in run_setup
          exec(code, locals())
        File "<string>", line 65, in <module>
[...]

PR #552, opened 2025-03-24
Fix setup.cfg metadata

h-mayorquin:
Hey guys, specifying metadata with dashes is deprecated on the last version of setuptools:

https://setuptools.pypa.io/en/stable/history.html?utm_source=chatgpt.com#bugfixes

This is causing people that want to build your package using the latest version of setuptools to fail:

https://github.com/NeuralEnsemble/python-neo/pull/1660#issuecomment-2749208250

This PR should fix it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,7 +10,7 @@ all_files  = 1
 upload-dir = build/docs
 
 [metadata]
-description-file = README.rst
+description_file = README.rst
 
 [aliases]
 test=pytest


h-mayorquin:
This should fix #551 I think.

jgrewe (approved):

jgrewe:
@h-mayorquin thank you

h-mayorquin:
Thanks for the merge.

Btw, reading about this yesterday I realized that `description_file` is not even a valid metadata field. I think the correct field is `long_description`. Apparently, someone used it at some point and people have been copying it from other setup-like configurations but the correct 

I suggest a move to pyproject.toml.

PR #553, opened 2025-05-16
Moving the project to a pyproject.toml file

wendtalexander:
I don't know how to test this properly it works on my machine simply with 

~~~sh
pip install .
~~~

in the project directory.

you can now specify which dependencies you want to install. If you want all dependencies for testing and building the docs you can install them with this command 
~~~sh
pip install ".[all]"
~~~


wendtalexander:
With the pyproject.toml file the setup.py file is obsolete and I removed it, but appveyor, and the github actions  expecting a setup.py, and I dont know how to fix these things. 

Maybe we can replace this with 
~~~sh
python -m build . 
~~~
or 
~~~sh 
pip install ".[all]"
~~~

And maybe can we drop support for python 3.8, because its end of life, and the setuptools required for python3.8 does not support the license-files key in the pyproject.toml file, which is now standard.

wendtalexander:
I moved the optional-dependencies to a single [dev] key in the pyproject.toml file so we can install all dependencies with one call 

~~~python
pip install ".[dev]"
~~~

this can replace the 
~~~python 
python setup.py install 
~~~
in all the GitHub actions that use this line.

jgrewe (approved):
LGTM

